PR #16, opened 2016-11-27
Onetruelit

chick:
A whole bunch of changes to make dsptools "one true literal" compliant

### DIFF
--- a/src/main/scala/dsptools/counters/Counters.scala
+++ b/src/main/scala/dsptools/counters/Counters.scala
@@ -66,21 +66,21 @@ class CountIO (countParams: CountParams) extends Bundle {
   // Count up/down control signal
   val upDown = if (countParams.countType == UpDown) Some(Input(Bool())) else None
   // Counters usually increment by 1
-  val inc = if (countParams.incMax != 1) Some(Input(UInt(countParams.incMax))) else None
+  val inc = if (countParams.incMax != 1) Some(Input(UInt(countParams.incMax.W))) else None
   // Counter wrap to value (up counters default wrap to 0)
-  val wrapTo =  if (countParams.customWrap) Some(Input(UInt(countParams.countMax))) else None
+  val wrapTo =  if (countParams.customWrap) Some(Input(UInt(countParams.countMax.W))) else None
   // Counter default wrap condition is when count is maxed out (so need to know max)
   val max = {
     if (countParams.wrapCtrl == Internal && countParams.countType != UpMod) {
-      Some(Input(UInt(countParams.countMax)))
+      Some(Input(UInt(countParams.countMax.W)))
     }
     else {
       None
     }
   }
   // n in x%n
-  val modN = if (countParams.countType == UpMod) Some(Input(UInt(countParams.countMax+1))) else None
-  val out  = Output(UInt(countParams.countMax))
+  val modN = if (countParams.countType == UpMod) Some(Input(UInt((countParams.countMax+1).W))) else None
+  val out  = Output(UInt(countParams.countMax.W))
 }
 
 /** Counter template */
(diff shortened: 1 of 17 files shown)


src/main/scala/dsptools/counters/Counters.scala
```diff
@@ -66,21 +66,21 @@ class CountIO (countParams: CountParams) extends Bundle {
   // Count up/down control signal
   val upDown = if (countParams.countType == UpDown) Some(Input(Bool())) else None
   // Counters usually increment by 1
-  val inc = if (countParams.incMax != 1) Some(Input(UInt(countParams.incMax))) else None
```
ducky64: This seems like a literal rather than a width?
Though the original code may have been just wrong, since you shouldn't have a literal inside an input...

Similarly for the next few instances.

src/main/scala/dsptools/counters/Counters.scala
```diff
@@ -164,11 +165,13 @@ abstract class Counter(countParams: CountParams) extends Module {
       //    else Mux(wrap,wrapTo,nextInSeq)
     }
     else {
-      if(wrap == Bool(true)) {
-        wrapTo
-      } else {
-        nextInSeq
+      val nextThing = Wire(Bool())
```
ducky64: Possibly written more concisely with a mux?

src/test/scala/examples/Demod.scala
```diff
@@ -43,12 +43,12 @@ class Demod[T <: Data:Ring](gen: => T, p: DemodParams) extends Module {
     // People aren't generally consistent about choosing positive LLRs to correspond to 0 or 1, so we choose
     // one with a convenient interpretation in this context.
     // Offset of the input sample relative to frame size (needs to support up to max frame size)
-    val offsetIn = Input(UInt(p.frameSizes.max-1))
+    val offsetIn = Input((p.frameSizes.max-1).U)
```
ducky64: Does it make sense to have a literal in an Input?

src/main/scala/dsptools/counters/Counters.scala
shunshou: A hardware mux is different from compile-time configuration, even if you can assume optimizers down the line will do the same thing [b/c Mux select is a Lit].

src/main/scala/dsptools/counters/Counters.scala
shunshou: Actually, sorry. My comment was referring to my original code, with DSPUInt[max]. Seems like when Chick translated it over to Chisel3, it got incorrectly replaced with UInt[Lit]. 

Original being `val inc = if (countParams.incMax != 1) Some(DSPUInt(INPUT,countParams.incMax)) else None`

src/main/scala/dsptools/counters/Counters.scala
ducky64: Is this a compile-time optimization? There's a comparison between Chisel types (between `wrap` and `true.B` - though in the original there's a double-equals `==` Scala-land compare, which I'm unsure will work correctly) which would look like you intended to generate a hardware compare?

src/main/scala/dsptools/counters/Counters.scala
ducky64: For automatically removing literals from IO Bundles: unclear whether this is a good idea. In ucb-bar/chisel3#168 rocket uses the same notation, but using the literal as a type (shorthand for not re-writing the bitwidth).

shunshou: Again, in Chisel2, I never wanted IO Bundles. I just wanted to be able to separately specify whether each signal was an IO or not, so I hacked my way around it, because no one wanted to support it.

src/main/scala/dsptools/counters/Counters.scala
shunshou: Ok, sorry. Again didn't pay attention to Chick's changes. My original code was 

```
val up = {
    if (countParams.incMax == 1 || (countParams.wrapCtrl == External && countParams.customWrap))
      (io.out + inc).shorten(countParams.countMax)
    else upCustom
  }
```

Which is purely compile time.

shunshou: [My] Original code was 
```
val nextCount = {
    if (countParams.wrapCtrl == Internal && (countParams.countType == UpMod || 
       (countParams.countType == Up && countParams.incMax > 1 && !countParams.customWrap))) 
      nextInSeq
    else Mux(wrap,wrapTo,nextInSeq)
  }
```

Which is compile time-configurable.

Anyways, didn't code review the changes made. My original counter, as verbose as it was, should have been functionally correct and runtime/compile-time configurable.

chick:
I have replace my complicated code with Mux per @shunshou, but not sure what to do about the other two comment threads.  Also would be nice if any of these things could be caught by some test. Everything changed here made it through ```sbt test``` successfully regardless of which way I did them

src/test/scala/examples/Demod.scala
shunshou: Should be log2up

PR #24, opened 2016-12-15
Invert reset

chick:
This is not ready for merge but it is ready for discussion.
Currently requires top level module to annotate itself
```
class ExampleModuleNeedsResetInverted extends Module with ResetInverter {
  val io = IO(new Bundle {
    val out = Output(UInt(32.W))
  })

  val r = RegInit(0.U)

  invert(this)
}
```
This may not be the right way to turn it on. 
Questions:
 * How to turn on.
 * Modules may turn this on for when they are top during testing, so maybe that should be ignored if they become subs later on


src/test/scala/dsptools/ResetInverterSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+
+package dsptools
+
+import chisel3._
+import chisel3.util.RegInit
+import dsptools.firrtltools.ResetInverter
+import firrtl._
+import org.scalatest.{FreeSpec, Matchers}
+
+class Unit extends Module with ResetInverter {
```
grebe: Unit is probably not a good name given it's meaning in scala

src/main/scala/dsptools/firrtltools/ResetInverter.scala
```diff
@@ -0,0 +1,66 @@
...
+        CircuitState(ResetN.run(state.circuit), LowForm)
+      case annotations =>
+        throw DspException(s"There should be only one InvertReset annotation: got ${annotations.mkString(" -- ")}")
+    }
+  }
+}
+
+trait ResetInverter {
+  self: chisel3.Module =>
+  def invert(component: InstanceId): Unit = {
```
grebe: I think it is better if mixing the trait in always causes invert to be called. I think it is legal to add invert(this) to the end of the trait.

src/test/scala/dsptools/ResetInverterSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  val r = RegInit(0.U)
+
+  invert(this)
+}
+
+class ResetNSpec extends FreeSpec with Matchers {
+
+  "Inverting reset needs to be done throughout module" in {
+    val optionsManager = new ExecutionOptionsManager("dsptools") with HasChiselExecutionOptions with HasFirrtlOptions {
```
grebe: I just want to be sure I understand correctly- is the transformation we are running getting passed in via the annotation? Nothing in the test is being done to build a custom firrtl compiler? If so, that is extremely cool.

grebe:
What are all the reasons for inverting reset? As far as I'm aware, the only reasons are integrating external IP and obeying standard interfaces for verification (e.g. AXI). What is our philosophy in such cases? Do we want to convert our default reset to an active low reset at axi boundaries, or do we want the entire circuit to be reset low?

I guess one approach I can imagine working is something where you mixin a ResetActiveLow trait or a ResetActiveHigh trait if you care about your reset for some reason, and otherwise you just inherit your parent's reset. I find the idea that you would conditionally invert your reset based on whether or not you are parent or child confusing- why would that be?

I am not sure what all the asic tool implications are for having reset low and reset high mixed together- don't we use something similar to clock tree synthesis for resets? Will this play nice?

stevobailey:
I think another reason just has to do with standard cell design. Active low resets take less logic I think. Though I don't think it makes a big enough difference to matter for our work.

I find it rather scary to "just inherit your parent's reset". I suppose if you don't do anything manually with the reset signal, it's fine. But if you have some IP block deep inside your hierarchy, you could unknowing change its reset polarity by changing it higher up.

The ASIC tools handle reset polarity automatically. Since these resets are synchronous, they actually treat them like logic I believe. I wouldn't worry about mixed reset polarities here.

My inclination is to first make it everywhere active high or active low. If that doesn't meet your needs for some reason, then mixing in a trait seems okay, though I'm worried about a hidden polarity. A more verbose albeit explicit solution would be active high or low everywhere, but a polarity specification per register if you actually care. Thoughts?

grebe:
Yeah, I think anywhere you have a blackbox you should specify reset polarity explicitly (or at least have a default active high that you can explicitly override). Blackboxes definitely shouldn't inherit from parents. But as long as you stay in chisel-land, the polarity of the reset doesn't really matter until you get out of chisel-land via verification tools (or something).

chick:
@grebe @stevobailey @azidar This is languishing.  It was a high priority todo at one point, as such it must have had some spec written down that we can look at and decide if this is done or not.

stevobailey:
If we included Paul's suggestion of inverting within the trait rather than the module, I'm happy to call it done and merge. It was mostly high priority because of outside pressure, I believe, so perhaps we should merge and wait for outside feedback if it arrives.

One comment, though. The test seems to just check the firrtl for a "not" gate on the reset and the wire to be renamed "reset_n". Can we have a test that actually tries to reset some register and verify the polarity that way? The hardware engineer in me is concerned by a test that just checks existence of some thing and doesn't check an example application.

PR #25, opened 2016-12-15
Remove clock and reset from floating point models

grebe:
Is this OK? It passed tests for me.

I have noticed (and so have some NGC folks) that some tools complain about ports not being connected when using the floating point blackboxes b/c the clock and reset don't get connected. I removed them and the tests still pass. Is it required anywhere that blackboxes have clock and reset?

@stevobailey may be interested, too

### DIFF
--- a/src/main/resources/BlackBoxFloat.v
+++ b/src/main/resources/BlackBoxFloat.v
@@ -1,6 +1,4 @@
 module BBFFromInt(
-    input  [0:0] clk,
-    input  [0:0] reset,
     input  [63:0] in,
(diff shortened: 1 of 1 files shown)


chick:
I don't know if it matters or not, maybe @ducky64 or @azidar can opine.  I've tested locally I can't see any difference in test behavior

grebe:
I forget the exact context of the errors, when I hunted them down it was just because the emitted verilog didn't connect the clock and reset bins and whatever tool it was (verilog maybe?) was complaining about it.

PR #26, opened 2016-12-16
[stevo]: support negative numbers in conversion from double to BigInt…

stevobailey:
…Bits

A really ugly solution, but it seems to be working for now. Thoughts on a better approach?

### DIFF
--- a/src/main/scala/dsptools/Utilities.scala
+++ b/src/main/scala/dsptools/Utilities.scala
@@ -33,7 +33,11 @@ object Utilities {
   def isEven[T <: Data:Ring](number: T, trim: TrimType = Truncate): Bool = ! isEven(number, trim)
 
   def doubleToBigIntBits(double: Double): BigInt = {
-    BigInt(java.lang.Double.doubleToLongBits(double))
+    if (double < 0) {
+      BigInt(java.lang.Double.doubleToLongBits(-double))+BigInt(1<<30)*BigInt(1<<30)*BigInt(8)
+    } else {
+      BigInt(java.lang.Double.doubleToLongBits(double))
+    }
   }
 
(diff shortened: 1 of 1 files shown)


src/main/scala/dsptools/Utilities.scala
```diff
@@ -33,7 +33,11 @@ object Utilities {
   def isEven[T <: Data:Ring](number: T, trim: TrimType = Truncate): Bool = ! isEven(number, trim)
 
   def doubleToBigIntBits(double: Double): BigInt = {
-    BigInt(java.lang.Double.doubleToLongBits(double))
+    if (double < 0) {
+      BigInt(java.lang.Double.doubleToLongBits(-double))+BigInt(1<<30)*BigInt(1<<30)*BigInt(8)
```
grebe: I think you can also give bigint a string that it parses, which would probably be cleaner

shunshou:
You should just be doing what chisel2 / my old tester did. Rather than some
arbitrary shift by 30...

stevobailey:
What did chisel2 / your old tester do?

shunshou:
Don't you need to first define # of fractional bits?

stevobailey:
Yes, this version is only valid for floating point at the moment. I have a larger item on my todo list of supporting fixed point for some of these operations, but we're running into chisel bugs on that end.

shunshou:
Ok well 1) I think getting fixed point to work *is* a pressing issue. 2) You did what Chisel testers did. At some point, Chisel2 became sign agnostic, and you'd reinterpret the sign only after reading generated values again, granted, their methodology has changed in the last ~half year so idk. 

https://github.com/ucb-bar/chisel/blob/master/src/main/scala/Tester.scala

PR #27, opened 2016-12-16
[stevo]: add axiRead, modify axiWrite slightly

stevobailey:
it's working in the template. able to read and write control registers, and read status registers.


stevobailey:
Seems to be working. It'd be great to have a test for this eventually.

PR #31, opened 2017-01-12
Repaired fixed point problems

chick:
This makes ```sbt test``` run successfully again.
Should probably be combined with [firrtl-interpreter PR #67](https://github.com/ucb-bar/firrtl-interpreter/pull/67)

### DIFF
--- a/src/main/scala/dsptools/DspTester.scala
+++ b/src/main/scala/dsptools/DspTester.scala
@@ -95,7 +95,7 @@ class DspTester[T <: Module](c: T,
 
   // [stevo]: poke a value in type typ to a UInt input
   //scalastyle:off cyclomatic.complexity
-  def dspPokeAs[TT<:Data](bundle: Data, value: Double, typ: TT): Unit = {
+  def dspPokeAs[U<:Data](bundle: Data, value: Double, typ: U): Unit = {
     bundle match {
       case u: UInt =>
         typ match {
@@ -174,7 +174,7 @@ class DspTester[T <: Module](c: T,
   }
 
   // [stevo]: peek a UInt but cast it to another type
-  def dspPeekAs[TT<:Data](data: Data, typ: TT): Either[Double, Complex] = {
+  def dspPeekAs[U<:Data](data: Data, typ: U): Either[Double, Complex] = {
     data match {
       case u: UInt =>
         typ match {
@@ -244,7 +244,7 @@ class DspTester[T <: Module](c: T,
 
   def nearlyEqual(a: Double, b: Double): Boolean = {
     val epsilon = 1e-12
-    val divisor: Double = if(b == 0.0) b.abs else epsilon
+    val divisor: Double = if(b.abs < epsilon) epsilon else b.abs
     val result = (a - b).abs / (divisor + epsilon) < epsilon
     //if(!result) {
     //  println(f"Got error on a $a vs $b  ${a-b}%20.16f")
--- a/src/main/scala/dsptools/DspTester.scala
+++ b/src/main/scala/dsptools/DspTester.scala
@@ -245,7 +245,11 @@ class DspTester[T <: Module](c: T,
   def nearlyEqual(a: Double, b: Double): Boolean = {
     val epsilon = 1e-12
     val divisor: Double = if(b == 0.0) b.abs else epsilon
-    (a - b).abs / divisor < epsilon
+    val result = (a - b).abs / (divisor + epsilon) < epsilon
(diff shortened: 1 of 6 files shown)


src/main/scala/dsptools/DspTester.scala
```diff
@@ -95,25 +95,29 @@ class DspTester[T <: Module](c: T,
 
   // [stevo]: poke a value in type typ to a UInt input
   //scalastyle:off cyclomatic.complexity
-  def dspPokeAs[T<:Data](bundle: Data, value: Double, typ: T): Unit = {
+  def dspPokeAs[TT<:Data](bundle: Data, value: Double, typ: TT): Unit = {
```
grebe: Maybe instead of `TT` (which to me implies some sort of connection with `T` we should go with `U` or `A` or something

src/main/scala/dsptools/DspTester.scala
```diff
@@ -238,12 +242,22 @@ class DspTester[T <: Module](c: T,
     }
   }
 
+  def nearlyEqual(a: Double, b: Double): Boolean = {
+    val epsilon = 1e-12
+    val divisor: Double = if(b == 0.0) b.abs else epsilon
+    val result = (a - b).abs / (divisor + epsilon) < epsilon
```
grebe: I don't think the `+ epsilon` is needed if the previous line ensures divisor =/= 0

src/main/scala/dsptools/DspTester.scala
```diff
@@ -238,12 +242,22 @@ class DspTester[T <: Module](c: T,
     }
   }
 
+  def nearlyEqual(a: Double, b: Double): Boolean = {
+    val epsilon = 1e-12
+    val divisor: Double = if(b == 0.0) b.abs else epsilon
```
grebe: I just realized this is wrong. First, I think you want to switch the true and false statements, and second of all I think it would be better to do:
`if (b.abs < epsilon) epsilon else b.asb`

src/test/scala/dsptools/numbers/BlackBoxFloat.scala
```diff
@@ -2,47 +2,84 @@
...
 
 //scalastyle:off magic.number
 
 class BlackBoxFloatTester extends BasicTester {
-  val (cnt, _) = Counter(Bool(true), 10)
+  val (cnt, _) = Counter(true.B, 10)
   val accum = Reg(init=Wire(DspReal(1.0)))
 
-  val addOut = accum + DspReal(1.0)
+  private val addOut = accum + DspReal(1.0)
```
grebe: Does this need to be private?

src/test/scala/dsptools/numbers/FixedPrecisionChangerSpec.scala
```diff
@@ -45,34 +45,34 @@ class RemoveMantissaTester(c: RemoveMantissa, inValue: Double, outValue: Double)
 class FixedPrecisionChangerSpec extends FreeSpec with Matchers {
   "assignment of numbers with differing binary points seems to work as I would expect" - {
     "here we assign to a F8.1 from a F8.3" in {
-      chisel3.iotesters.Driver(() => new FixedPrecisionChanger(8, 3, 8, 1)) { c=>
+      dsptools.Driver.execute(() => new FixedPrecisionChanger(8, 3, 8, 1), Array.empty[String]) { c =>
```
grebe: Does the dsptools driver not set the second argument to Array.empty by default?

grebe: Also, will these tests not work if you use the iotesters.Driver? What makes dsptools Driver different?

shunshou: Your last question was exactly why I wanted clarification about different drivers...

chick: dsptools.Driver adds the fixed point black boxes to the options so they show up for the interpreter

chick:
@grebe @shunshou : the dsptools.Driver.execute automatically includes the scala implementations of the fixed point black boxes for the interpreter.

shunshou:
I think, then, you should mention in documentation somewhere that dsptools.driver supersedes chisel.iotesters for anything dsp related [so you can have reals]. When looking at different examples before, I think there were sometimes inconsistent use cases (maybe reals weren't used for a particular example, and therefore it was ok to use iotesters?).

Issue #30, opened 2017-01-12
Support for trig functions w/ Verilator


shunshou:
I believe with SystemVerilog, you can do stuff like $atan, so I assume this should work with Verilator? I'd like to be able to compute an angle with CORDIC, which only operates on synthesizable bits, but also be able to swap the CORDIC block out with some non-synthesizable trig function + appropriate delays, for the purposes of simulating SQNR/performance degradation between floating pt and fixed pt types in a full system**. I can get around having it in dsptools by ExtModuling a BB Verilog wrapper, which might make sense... but at the same time, it might make more sense to wrap CORDIC in a function with params that can be switched off with a non-synthesizable implementation. 

And just as a reference, apparently trig stuff was supported in Chisel2 Flo.

shunshou:
In general:

Trig functions for Real #'s
DSPTester with Verilog dump
Documentation of ExecutionOptions
RoundHalfUp, Floor [at the minimum]
Ranges for control logic [UInts mostly]
Need to double check, but is ROM support in Chisel3 nice like in Chisel2? Or is it just a nasty bunch of logic? I'd like to use a ROM compiler instead of logic for easier place & route, so this should be special cased out.
Multiclock [divided down--should the divider be implemented digitally for timing mismatch reasons?], clk IO

PR #33, opened 2017-01-13
Add more floating point ops and more tests.

grebe:
Initial pass addressing #30

The tests aren't currently passing. @chick, can you see what I'm doing wrong? I think I'm invoking the driver correctly. I think dsptools.Driver doesn't currently include the verilog file (it probably should), but it seems to me that the driver should work with the firrtl backend.


shunshou:
Curious: is there no round and just floor/ceil?

grebe:
@shunshou not in IEEE.1364-2005 that I can see

shunshou:
Also, don't these functions fall out of your general Ring, whatever whatever types? In spire, there's something specific for trig, but you should prob also think harder (eventually) on how you want to categorize stuff...

grebe:
These are only defined for DspReal, not in any of the typeclasses. Spire and Breeze have typeclasses for these things, but they aren't easily transferable to hardware because they rely on comparisons and stuff (comparisons exist in my typeclasses, but are different b/c you don't get the result instantly). Long story short, I think trig operations probably don't really make sense in a typeclass because they are probably expensive enough that you want to think about them very carefully. I think division also probably doesn't belong, which is why I use Ring but don't really use Field

shunshou:
Ok that's fair.

shunshou:
@chick or someone who knows better, can we try to get this up ASAP? Really important for testing my blocks at a systems level...

chick:
@grebe, the first errors I see are a bunch of errors of the form
```
_BlackBoxFloat.v:264: Unsupported or unknown PLI call: $atanh
```
It looks like BlackBox.v is being seen by verilator but it is stuff in there that is not supported.
I have tried replacing all the reported errors $pos => pos etc and this part of the tests passes.  Does my solution to this part seem ok

chick:
I have pushed a fix that improves things.  The verilator problems with ```pos```, ```atanh``` etc is still outstanding, my remove dollar signs changed error but did not fix it.

OTOH, I have fixed most of the interpreter problems with the new operators, the DspFactory that resolves references had spaces a the end of all the new operator names.  I have fixed this and will create an issue with interpreter that is should not silently ignore missing black boxes,

grebe:
Verilator still doesn't find the blackboxes for me- am I on the wrong branch? I think my chisel-testers is @ 22f0c37036cd396c64a5b947f54b3d68f4707ea4 (Wed Dec. 11)

grebe:
The interpreter does work, though, which is excellent

src/main/scala/dsptools/numbers/DspRealBlackBox.scala
```diff
@@ -266,27 +266,27 @@ class DspRealFactory extends BlackBoxFactory {
       case "BBFFromInt"           => Some(add(new DspRealFromInt(instanceName)))
       case "BBFToInt"             => Some(add(new DspRealToInt(instanceName)))
       case "BBFIntPart"           => Some(add(new DspRealIntPart(instanceName)))
-      case "BBFLn "               => Some(add(new DspRealLn(instanceName)))
```
grebe: This was a very silly mistake- good catch!

shunshou:
I'm semi-strictly working with Verilator b/c I'm running with non-standard clks, so it'd be preferably to get this to work w/ Verilator. It's possible that Verilator just doesn't supper atanh and stuff? Hopefully the subset of things that do work (if any) meet my needs......

grebe:
@chick can we merge or are we waiting on something? I don't think we should wait for any verilator patches or anything, it's not as easy as I thought it was.

chick:
merged

PR #37, opened 2017-01-15
Tester updates

stevobailey:
Note that "toBigInt" and "toDouble" are defined in the Chisel3 FixedPoint object. So I removed them. Here I create new versions that map to and from positive BigInts (assuming the sign bit is the MSB). I call it "unsigned" but it's really a positive mapping of the signed integer. I've also expanded the packing and unpacking tester functions to include FixedPoint and Complex types. SInt stuff hasn't been tested, but the rest has.

### DIFF
--- a/dspblocks/src/main/scala/dsptools/DspBlock.scala
+++ b/dspblocks/src/main/scala/dsptools/DspBlock.scala
@@ -162,7 +162,7 @@ abstract class DspBlockTester[V <: DspBlock](dut: V, maxWait: Int = 100)(implici
         f.asInstanceOf[FixedPoint].binaryPoint match {
           case KnownBinaryPoint(binaryPoint) =>
             in.map(x => x.reverse.foldLeft(BigInt(0)) { case (bi, dbl) => 
-              val new_bi = toBigInt(dbl, binaryPoint)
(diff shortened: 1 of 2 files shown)


dspblocks/src/main/scala/dsptools/DspBlock.scala
```diff
@@ -176,43 +178,126 @@ abstract class DspBlockTester[V <: DspBlock](dut: V, maxWait: Int = 100)(implici
...
         val epsilon = 1e-12
-        val err = abs(c-r)/abs(r+epsilon)
+        val err = abs(c-r)/(abs(r)+epsilon)
         assert(err < epsilon || r < epsilon, s"Error: mismatch on output $index of ${err*100}%\n\tReference: $r\n\tChisel:    $c")
       }
     }
   }
 
+  // compares chisel and reference outputs, errors if they differ by more than epsilon
+  def compareOutputComplex(chisel: Seq[Complex], ref: Seq[Complex]): Unit = {
```
stevobailey: Note that Seq[Double] and Seq[Complex] both reduce to Seq, so if we try to overload one method with the other, Scala complains.

stevobailey:
Merging to keep dsp-framework on the master branch as I update dependent repos and push them to the chamber.

PR #41, opened 2017-01-16
Add shift register with reset

stevobailey:


shunshou:
should really be added to Chisel utils... not dsptools...

Issue #17, opened 2016-11-29
simplify generic type usage


stevobailey:
Change this:
Wire(implicitly[Real[T]].zero)

into this (possibly still need the Wire wrap):
Real[T].zero

stevobailey:
bump

PR #44, opened 2017-01-16
PML double2T 

### LINKED ISSUE #36: How to do Double to T conversion?
if I have gen : => R, where R <: Data:Real, how do I get a constant (say 3.3333) of type R? 

I would have wanted to try something like gen.fromDouble(3.3333), but I got `value fromDouble is not a member of type parameter R`. 

I saw in Stevo's FFT: 

`implicitly[Real[R]].fromDouble(3.333333)`

At least this compiles...

But my question then is, how do you specify the width, # of fractional bits of the constant?

I assume, through some kind of implicit something or another, if gen was passed FixedPoint(32.W, 16.W), it'd know that the constant would have width = 32, with 16 bit fractional width...

But if I wanted to have another constant with width = 16, fracWidth = 8, would I need another gen? i.e.

gen1 : => R for 32,16
gen2 : => A for 16,8

?

Also, why does gen1.fromDouble / gen2.fromDouble not work? And instead I have to use implicitly[Real[R]].fromDouble, implicitly[Real[A]].fromDouble?

Or rather, is the width just supposed to be auto-inferred somehow? 

I noticed that in Chisel3, for FixedPoint, you have:

`def fromDouble(value: Double, dummy: PrivateType = PrivateObject,
                 width: Int = -1, binaryPoint: Int = 0)`

But the ability to specify width and binary point doesn't seem to be carried over to the fromDouble in ConvertableToFixedPoint in dspTools?

These kinds of things *really really really* should be documented...

shunshou:
Temporary solution to not being able to specify binary pt for type-parameterized constants. 
Warning: I do Round(dbl).toInt instead of dbl.toInt

Fixed #36 and somewhat related #17 (?)


shunshou:
Going to merge b/c it's gating me... @grebe and @chick should come up with a better soln at some pt.

Issue #38, opened 2017-01-16
DSPTester peek/poke printing more useful info


shunshou:
I think you shouldn't directly be using TesterOptionsManager and should have a DSPTesterOptionsManager that lets you customize verbosity.

I finally got the standard PeekPokeTester to print out something, but evidently, things are just printed out as BigInt converted to base 10. 

My old DSPTester printout allowed for much more thorough debugging:

You could print out base 10/hex bits flavor. It would also directly print out the decimal version of DspReal/FixedPoint peek/pokes... and for fixed point, it would also show # integer bits + # fractional bits used (or it might've been range, I forget) to compute the Double result for fixed point (that helps to double check that width inference, etc. were working correctly). 

All of these features help to make the debug process easier.

For example: If I want to test the correctness of `implicitly[Real[R]].fromDouble(3.333333)`

And I'm not super sure how many Fixed point fractional/integer bits are allocated, I'd like to use the tester to debug... 

I'm also not loving the fact that expect prints twice (the first line is ugly and non-informative -- you don't even know which signal it's referring to):

expect got      2.00000000 expect      2.00000000
EXPECT AT 15 add 1.0 + 1.0 => 2.0 should have been 2.0 PASS

Expect should look more like the peek/poke strings, and if you supply a string, it should only print something like the 2nd line, but with a signal name [without you having to manually pass in result and expected vals]. 

Little details like this speed up the design/debug process considerably...

---

Edit: Ok, DspTester needs its own verbosity, so you can turn off _verbose for PeekPokeTester and override all the prints for Dsp specific things with something that's more useful...

shunshou:
I'm working on my own version of VerboseDspTester (extends DspTester) that gives me features I want. It should also make addressing #29 a piece of cake whenever I need to do it for hardware validation...

Issue #29, opened 2017-01-12
DSPTester printing Verilog


shunshou:
Forgot to bring this up at the meeting, but see:

https://github.com/ucb-art/ChiselDSP/blob/master/src/main/scala/Overlay/DSPTester.scala

Every peek/poke corresponds to equivalent Verilog TB syntax. Super useful for debugging stuff later down the chain (say whether memory swapping actually worked or not) when you're not using rocket-chip's build infrastructure.

grebe:
@shunshou can we close this issue? #46 was merged

PR #46, opened 2017-01-17
DSP Tester overlay with more features

shunshou:
[Do NOT merge]

Enhancements for #29 #38

@grebe can you double check that my removals of super.peek*'s in the last commit shouldn't screw anything up?


grebe:
lgtm

shunshou:
@grebe see last commit -- do you think double should be converted to int after rounding or just double.toInt for dsppoke where data is sint/complex r sint?

shunshou:
ack, same goes for complexpoke... didn't change that... although round might make more sense dspwise?

chick:
@shunshou why is the package of ```VerboseDspTesterUtils``` chisel3.iotesters? My style checker doesn't like it.

shunshou:
It doesn't need to be now. I needed it originally Bc the optionsManager was
private. BUT I needed to duplicate a peek on Aggregates in PeekPokeTester,
but that uses aggregate flatten, which is also only available to
chisel3.... That still needs to be in some chisel package unless chisel Api
guys are ok making it public (which I doubt????)

shunshou:
I realize the intention is to make everything "safe" but it's hard to add capability layers that might otherwise bloat chisel3 in a way people wouldn't be happy with when everything is privatized.

shunshou:
@grebe @chick can you guys briefly look over files that changed to see if it affects anything you guys expected? Probably don't bother with the files that I added... I'll add some tests for different peeks/pokes/expects, etc. just to double check things are working and then merge, hopefully sometime tomorrow.

src/main/scala/dsptools/Chisel3TestersCompatibility.scala
```diff
@@ -0,0 +1,14 @@
+package chisel3.iotesters
```
chick: @ducky64  @shunshou I think we should consider looking into what you need here and add it, if at all possible, to chisel3 API.  We are doing a lot of API cleanup right now, so it's a good time.  I see only one use of it, can you describe the use case

shunshou: Well, if you can make all of the methods I need in this file public, then I won't have an issue. I don't think it's unusual for people to want to override some feature with their own custom thing, so privatizing stuff is annoying. I'm largely adding extra print features, etc. to the tester that maybe people don't want in the base tester... but that requires overriding super class functions, and I don't always have access to the stuff used by the functions b/c I'm not in iotesters. I get that people should maybe not be trying this kind of thing on a regular basis (esp. for hardware writers) but for people trying to extend Chisel features via libraries, the privatizing thing is very limiting. Is there some way around that? And I don't mean just telling me that x,y,z is dangerous and that's why we're not supporting it (I feel like I've heard that way too much this past year) -- b/c I and others will find ways like this to go around stuff.

shunshou: @chick basically these are all things used by peek/poke, etc. in PeekPokeTester  that I had to recreate when I overrode stuff... The dspPeek, etc. in the original DspTester also didn't properly print out Poke signalName, so, just as in VerilatorBackend, I used getDataNames to get the right signalName (tho I need to double check things work). 

Flatten I needed for peeking aggregates @ the individual element level (but point to the right peek). 

etc. etc.

shunshou: Ok, I double checked. Pretty much all of the functions here are so that I can override a method that's like 90% identical to the old one with minor tweaks. (flatten, bigIntToStr, validName). validName was also used for dspPeek/Poke logger printing. 

I wanted getDataNames for being able to print a Verilog TB file just like how VerilatorBackend sets stuff up.

shunshou: Anyways, quick turnout to me = ~3 days max. Otherwise, I'll prob merge this with master, unless someone else has other complaints? It can be fixed later. It's not like this repo doesn't need a lot of refactoring to begin with... 

Maybe a better question to pose to @ducky64 is -- why privatize any of the stuff in my Compatibility file? I don't think it's possible that using these things will cause Chisel to break in any way. I get why addPin can be dangerous, but most of these things just add more visibility (like how a signal is actually specified in Verilog, etc.) that people can then use for some post processing (or in my case, Verilog TB creation...).

ducky64: So the goal of a good API is to both prevent unsafe things (like being to add pins at arbitrary points of elaboration) and to provide an intuitive and clean interface that makes sense and isn't brittle. Of course, the latter points are mostly pretty subjective...

- flatten: several reasons this is bad:
  - flatten doesn't make sense on Data, since Data can be either an Aggregate or Element. This is probably more a style issue, since the current implementation isn't incorrect. Though this is constraining in that everything must reduce to Bits, even where it might not make sense (Clock, or even other types we may not have thought of yet).
  - This also breaks individual levels of abstraction, for example you're essentially punching through several layers of Bundles and stuff to get the underlying elements. A better abstraction is to consider things on level at a time. This does require some structural changes, such as having hierarchical directionality (which is planned for the Bindings refactor).
  - Note: I want to deprecate this with the asTypeOf / Bindings refactor. Do you have a use case where you need it that can't be considered at each level (possibly recursively)?
- getDataNames, validName: these are tester internal details. There would be a good argument here on how to make testers extensible, and restructuring may make it possible for you to do what you need without this hack. Reason internal details are hidden is that they aren't part of the API, which gives developers flexibility to change the implementation without breaking user code.
- bigIntToStr: I don't even know why we have this. [BigInt.toString(radix: Int)](http://www.scala-lang.org/api/2.12.0/scala/math/BigInt.html#toString(radix:Int):String) exists.

shunshou: Well, I'd argue that currently, things are quite brittle and poor for extensibility. 

If you look, flatten is only functioning on Aggregates. It's just used to get the underlying elements of Aggregates so that you can peek at them one by one. It's necessary, b/c that's how the original PeekPokeTester implemented things, and I needed to change the function that was being called on the individual elements. An alternative would be to completely rewrite a function (that just recursively goes through elements), but that's what flatten gives you, so doing so seems incredibly redundant. 

IMO, API can be hidden (i.e. not in some kind of user doc), but still usable by "advanced users". If not, I'm literally going to take your function, copy and paste it somewhere, and remove "private" -- and you wonder why things get forked. 

When you ask for uses cases, to me, that implies more personal reasons for why I want to do X, Y, Z. This isn't even at that level of "use case." All I'm trying to do is change some small details in the way the original PeekPokeTester works that's more suitable for more informative testing of DSP without messing around with the original PeekPokeTester/DspTester (b/c doing that would make people even more unhappy, I presume). These functions were used in the original PeekPokeTester itself. If I want to change a tiny thing in my extended version of PeekPokeTester while keeping original functionality, I require a lot of copy/pasting of functions that are otherwise "private" to anything outside of PeekPokeTester. 

Re: bigIntToStr: Ask the person who first decided to use it in PeekPokeTester (or maybe it was VerilatorBackend). It's not always that straightforward for someone to realize that a function already exists unless they've seen it before... 

Anyways, can I expect a change to support these within the next n days, or should I just merge this into dsptools the way it currently is, since it seems like it's not as simple as just publicizing stuff? @chick Unless you have a better solution... 

---

Also, re: addPin, if I'm not the only one asking to keep the function in, then IMO, saying it's "unsafe" is not a good justification for taking it away.  The average person won't even see/know the function exists, but super users have reasons for using them, and it just serves to alienate them if you do remove such functionality without an acceptable backup in place. [If nothing else, it served as a workaround for features that didn't exist in Chisel...].

ducky64: Yeah, software engineering quality is a continuing work in progress. Hopefully things will get better.

flatten: yes, the implementation is correct, but ultimately it probably shouldn't be a member on Data. The alternative isn't necessarily to reimplement it, but to refactor or restructure use sites to use cleaner, more local methods, like associating directionality with Aggregates rather than inspecting the leaf elements. For example, PeekPokeTester could recursively pack / unpack bits into a Bundle, instead of flattening a blackbox Data to leaf Elements. From a structural standpoint, this would expose a packing / unpacking function that can have arbitrary implementation for each data type, rather than confining every data structure to be some collection of Elements.

As an example: what happens if you flatten something that includes an Analog type? Or how would you flatten an union type?

Hiding API: if it's part of the public API, someone is going to use it, and it makes it difficult to remove / change later if we discover it's bad. The closest thing we're doing is the experimental package, but ultimately every API item comes with an expectation of continued support. And people complain when things change, whether it was intended as a "public" API or not...

Obviously, the API we've exposed isn't perfect, but it would also be nice to work with us to make a better, more featured API in a principled way rather than punching holes into abstraction boundaries. There's usually a reason we've defined the abstraction boundaries in the way we did, and in a way that might not be immediately obvious to someone who isn't developing infrastructure.

For the PeekPokeTester API: I guess it could be structured better? But chisel-testers doesn't seem to see as much debate compared to the rest of core chisel. Much of the software engineering progress on core chisel has been because of the development effort by many, many people with different backgrounds and opinions.

addPin: you're right in that we should have had a replacement available when it was removed (I think this is what Record is for? Also keep in mind that the core Chisel team is limited so we can't implement everything everyone wants, but you're welcome to contribute =) ). Superusers and leaky abstractions also presents a problem: they will do something that pushes the boundaries of the system in a way the developers never intended, relying on internal implementation details such that continued development becomes impossible without changing externally visible behavior. Or they do things in a complex or clever way that ends up being as structurally sound as a house of cards. Ultimately there's a compromise between how fast internal development can happen, how many features are exposed, and how correct everything is. Chisel2 probably was pretty much in the "gradware" region of that compromise (expose everything, move fast and break things) while Chisel3 aims to be more stable and principled. However, that means not exposing everything like was done in Chisel2, which unfortunately appears to users as removal of features.

shunshou: I thought Record was originally to address the fact that I wanted Vecs with non-uniform widths? addPin was a totally separate thing. 

I get that Chisel is short-handed as far as developers are concerned. Unfortunately, "gradware" methodology probably worked best for me. As a Chisel user, waiting weeks/months/infinite time for a feature I want is kind of unrealistic, since, frankly speaking, I'm not paid by the hour, and needing Chisel with w/e features to be up to some certain standard (or just stable enough -- which it currently isn't) just so I can actually get my PhD (which isn't in tools + methodology) is really painfully frustrating. Case in point being memory stuff, which blew up in my face during the last tapeout. 

From your standpoint, it makes sense to support larger bodies rather than one off projects, simply b/c there just aren't enough man hours to support everyone within due time. If I actually had spare time and no urgent deadlines, then yeah, I wouldn't mind trying to work with people to find a better solution than punching holes. But if I'm supposed to be using this stuff to tapeout in <3 months, then that's not really viable, and to meet such a deadline, unless I get something like ~day support turnaround, hacking is really the only thing I can do. That's pretty much why dsptools was/is a mess. Just dump stuff where we can and deal with it later. From an actual hardware design risk mitigation standpoint, it makes more sense for me to be working in my Chisel2 framework, but that's not what the powers that be want me to do...

 [I can go into another long rant about how tapeout + infrastructure/tools development should be totally separate things and the former should not happen until the latter is on much more solid ground (even if the former kind of *should* inform the latter about how things should be done -- at least from a user experience POV), but that's... again... out of my control...]

shunshou: Also, why do you need to be the ones providing guarantees that "hacks" keep working? My opinion is:

Make a list of public functions that you know are correct and always will support.
Keep other stuff public as well to make library writing easier for people, but it's up to the library or w/e folks to make sure that their stuff doesn't break as Chisel changes. They have their own API, which should be even more front-facing if their libraries are being used, but that's not your concern.

ducky64: From a research standpoint, one of the goals was reusability of RTL-level components. While industry is better at engineering and building things that last (or maybe not, considering the widespread use of Perl to generate Verilog), if we had to rewrite Chisel every time it got too complex, then we really haven't accomplished easy re-use. And it would be a massive waste of time, not just for infrastructure builders, but for users who waste hours / days working through obscure bugs from complex unforeseen interactions.

There's also a trade-off between writing your one-of solution that basically needs to get thrown away, and being a part in building infrastructure. If everyone did the former, then everyone would be starting from near-scratch (since the last person's code doesn't work anymore / isn't structurally sound to build upon / isn't documented / is basically on fire) rather than being able to build on top of existing projects and systems.

If the tape-out deadlines are so tight that developers feel they are pressured into building hax on top of hax, that seems like a problem with the deadlines, possibly with misconceptions of how stable / how powerful the infrastructure is in the setting of deadlines. Infrastructure is (hopefully) the kind of stuff that gets build / fixed once and can be re-used in future projects, but hax upon hax are a recurring cost. And it usually gets worse each time.

shunshou: One of the goals of CRAFT, maybe. Not my goal in particular. I just want to build cool hardware. 

But I agree. In the perfect world, if everyone cared about reusability by other people (and not one-off ISSCC papers), then yes, more attention should be paid to infrastructure. It's just that deadlines are really out of our (or at least *my*) control. If people who had more control over these deadlines understood that people needed to nominally spend ~1 year on infrastructure before tapeout (and that it's kind of fruitless to tapeout while infrastructure is in its infancy), then we'd have a much more reasonable experience. Future tapeouts would involve less wastes of time rebuilding the same infrastructure over and over again. [Heck, that can be said about pretty much all of BWRC's "house of cards" infrastructure]. In the end, though, from many years of observation, a PhD is perceived(maybe wrongly so?) to be something *you* did in whatever field you're supposed to be in, so trying to be open source-y altruistic on something that's a tool for succeeding in your specialty unfortunately becomes somewhat tangential to that -- which kind of leads to this messy infrastructure -- but if no one else before you bothered cleaning up the mess, then is it really in *your* best interest to spend more time making things better for the people after you instead of getting papers in your name, etc. so you can graduate? It's a different kind of merit that makes more sense at a company or elsewhere (or maybe if your PhD specializes in operations management or something), where team success is your success (and keeps getting you a paycheck). But up until now, I haven't really seen this philosophy working in a PhD setting... [If that actually worked, my super tapeout wouldn't have been such a nightmare...]

For things that are easy to build upon, I'm totally fine helping out, but when I'm choosing between my own deliverables and the "general good" within a constrained time frame, it only makes sense that I focus on the former. I have a *bleep* ton of ideas for things I'd like to see in Chisel to be useful for *circuit designers* in the future; some of which I do plan on working on... but I don't want to fight the system to make it work. That's why I prefer layering on top of Chisel things via a library (which is wholly up to me or someone else at a higher-than-Chisel level to maintain), but it's just not easy to do if you don't have access to everything in Chisel. If I didn't care about infrastructure at all, I'd just be building all of my functionality into a "for Angie only repo", but that's not exactly how I've been working for the last year; otherwise, I probably would've been done  a lot sooner -- in that case, why care at all about user experience, right? Since I'm the only user. If nothing else, though, hacking a solution to a problem that hasn't been previously addressed by Chisel is still useful in the sense that at least people become aware that x, y, z feature can actually be useful. Then Chisel retainers or w/e can decide if it's useful enough to solidify and build into the core.

Anyways, all this talk about infrastructure and whatnot makes me think we should bring in some IEOR PhD student to figure out how to coordinate or get more industry people involved.

shunshou:
Don't worry. I have tests in a different repo, but will port them once they fully pass...

src/main/scala/dsptools/Driver.scala
```diff
@@ -8,23 +8,39 @@ import firrtl.{HasFirrtlOptions, ExecutionOptionsManager}
...
+import scala.util.DynamicVariable
+
 
 object Driver {
-  def execute[T <: Module](
+
+  private val optionsManagerVar = new DynamicVariable[Option[VerboseDspTesterOptionsManager]](None)
+  def optionsManager = optionsManagerVar.value.getOrElse(new VerboseDspTesterOptionsManager)
+
+  def execute[T <: Module, M <: TesterOptionsManager](
```
shunshou: Hm but it doesn't hurt to imply that it can be any subclass of TesterOptionsManager? But sure...

src/main/scala/dsptools/DspTester.scala
```diff
@@ -247,12 +247,11 @@ class DspTester[T <: Module](c: T,
...
   }
 
-  def dspExpect(data: DspComplex[_], expected: Complex, msg: String): Unit = {
+  def dspExpect(data: DspComplex[_], expected: Complex, msg: String): Boolean = {
     dspPeek(data) match {
       case Right(complex) =>
         println(f"expect got $complex expect $expected")
-        expect(nearlyEqual(complex.real, expected.real), msg)
-        expect(nearlyEqual(complex.imag, expected.imag), msg)
+        expect(nearlyEqual(complex.real, expected.real), msg) & expect(nearlyEqual(complex.imag, expected.imag), msg)
```
shunshou: lol I guess I use & and && interchangeably in a bad way... since it works with Booleans...

shunshou: Actually I think it's & in this case?

grebe: It shouldn't be. & is bitwise-and, && is logical-and. You're returning a value that represents an outcome, right? true->pass, false->fail? That should be logical-and. For boolean it works out to being the same, but it's just a little confusing b/c with & you always have to go check the types of things you're and-ing.

src/main/scala/dsptools/VerboseDspTester.scala
```diff
@@ -0,0 +1,303 @@
...
+    }
+    if (isSigned(signal)) signFix else bigInt
+  }
+
+  ///////////////// OVERRIDE UNDERLYING FUNCTIONS FROM PEEK POKE TESTER /////////////////
+
+  override def step(n: Int) {
+    if (dispDSP) logger println s"STEP ${n}x -> ${t+n}"
+    stepPrint(n)
+    backend.step(n)
```
shunshou: The problem is I can't change _verbose in PeekPokeTester easily. Unless I mess around with the DynamicVariable inside optionsManager.testerOptions.isVerbose. Even then, I have an alternative purpose for that particular isVerbose (i.e. print out everything bit-wise, rather than as Double for dspPeeks, etc.). The backend versions of stuff lets me directly pass in verbose as an argument. If someone wants to provide an alternative, that's fine... but again, I'm just really going around restrictions of chisel-testers...

src/main/scala/dsptools/VerboseDspTesterUtils.scala
```diff
@@ -0,0 +1,135 @@
+package dsptools
+
+import chisel3._
+import java.io.{File, FileWriter, BufferedWriter}
+import chisel3.iotesters.TestersCompatibility
+
+// Note: This will dump as long as genVerilogTb is true (even if you're peeking/poking DspReal)
+trait VerilogTbDump {
```
shunshou: Mixing traits like that doesn't seem to work. I can't access stuff like incTime, etc. from the trait.

grebe: Are you using a self type?

shunshou: Nope. I tried earlier for a different reason. I couldn't figure out how to use it properly. But overriding step/reset serves other purposes [and overriding any functions in general].

1) Verbose = make the logger print more useful
2) Dumps to a TB file. 

I'd rather keep only stuff that dumps to a TB file in the trait.

PR #50, opened 2017-01-19
Work to make the IsReal family work for FixedPoint

chick:
implements ceil, floor, isWhole, round, does not do toDouble for now
Fix declaration of isWhole in Ops.scala IsRealOps
Tests for above
Maded ParameterizedAdderTester test negative numbers too


shunshou:
I think it's dangerous to do toDouble. Thanks for working on this! :) I'll take a look once I'm done with some GSI stuff...

shunshou:
I take that back regarding to double. It'd be cool to do it so the golden model can be written essentially in scala and certain blocks can then be swapped out for fixed point versions to see how the system functionality degrades. Anyways, useful feature for another time...

PR #59, opened 2017-01-25
Add initial travis scripts.

### LINKED ISSUE #43: Set up Travis CI
@chick @ucbjrl I don't have the permissions to set up travis. We should do this.

I'll add a .travis.yml if one of you don't mind setting up travis's side of things.

@ucbjrl is there a way to set it up to do nightly builds? I'd like to run a script that checks out the head of master of everything and check that it still works. Our last floating point regression shouldn't have been around for so long.

grebe:
Should close #43 after travis is enabled for this repo.

Travis now has a "cron jobs" feature that we should run every day for dsptools.

### DIFF
--- a/.install_deps.sh
+++ b/.install_deps.sh
@@ -1,0 +1,24 @@
+set -e
+set -x
+# Install Verilator (http://www.veripool.org/projects/verilator/wiki/Installing)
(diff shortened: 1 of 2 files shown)


chick:
lgtm;

grebe:
Let's wait to merge until travis is enabled

PR #60, opened 2017-01-26
Verbosetestercleanup

shunshou:
@grebe I've tested a bunch of test bench use cases + the generated Verilog/TB files. It's kind of cool b/c I can simulate simple designs via https://www.edaplayground.com (Aldec Riviera Pro doesn't require verification...). I'll commit my tests to this branch, but as a final step, I compare the tb.v generated with a known working version... that also happens to be ~800 lines long. I'm going to commit that big tb too (I guess for more thorough regression tests?), but if you think it's too much, I can remove the commit and try a smaller tb... lmk what you prefer & then I'll squash + merge.

### DIFF
--- a/src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
+++ b/src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
@@ -1,0 +1,401 @@
+package SimpleTB
+
+import chisel3._
+import chisel3.internal.firrtl.{Width, BinaryPoint}
+import chisel3.experimental.FixedPoint
+import chisel3.util.RegNext
+import breeze.math.Complex
+import dsptools.numbers.{Real, DspReal, DspComplex}
+import dsptools.numbers.implicits._
+import dsptools.{VerboseDspTester, VerboseDspTesterOptionsManager, VerboseDspTesterOptions}
+import org.scalatest.{FlatSpec, Matchers}
+import chisel3.iotesters.TesterOptions
+
+// TODO: Make utility!
+trait EasyPeekPoke {
+
+  self: VerboseDspTester[_] => 
+
+  def feed(d: Data, value: Double) = {
+    d match {
+      case b: Bool => poke(b, if (value == 0.0) 0 else 1)
+      case u: UInt => poke(u, math.abs(value.round).toInt)
+      case r => dspPoke(r, value)
+    }
+  }
+  def feed(d: DspComplex[_], value: Complex) = dspPoke(d, value)
+
+  // TB Debug only [otherwise, just use expect for any *real* tb!]
+  def checkP(d: Data, value: Double) {
+    d match {
+      case b: Bool => peek(b)
+      case u: UInt => peek(u)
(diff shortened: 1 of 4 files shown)


src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
```diff
@@ -0,0 +1,401 @@
...
+      var tbFileLoc: String = ""
+      dsptools.Driver.execute(() => new SimpleIOModule(p.genShortF, p.genLongF, includeR = false, p), 
+          optionsPassTB) { c => {
+        val tester = new PassIOTester(c)
+        tbFileLoc = tester.tbFileName
+        tester
+      } } should be (true)
+      def read(file: String) = scala.io.Source.fromFile(file).getLines.mkString("\n")
+      val tbTxt = read(tbFileLoc).stripMargin
+      require(tbTxt == TbGoldenModel.txt, "Test bench incorrect! (compared against simulated model)")
```
grebe: Might want to do something like: https://github.com/ucb-bar/firrtl/blob/master/src/test/scala/firrtlTests/AttachSpec.scala#L15, that way when there is an error you get more insight into where it was.

shunshou: Oh instead of a direct == ? Sure.

shunshou: I would argue something like 

TbGoldenModelTxt.zip(tbTxt) foreach { case (expected, in) =>
        (expected == in) should be (true)
      }

is better? Rather than contains?

grebe: It just depends on how brittle you want the test to be. I guess your testbench is very sequential so contains won't catch all problems, but you might not want the test to fail if you move / change comments.

shunshou: Also
```
should properly poke/peek io delayed 1 cycle with gen = fixed + print TB (no reals) *** FAILED ***
[info]   false was not true (SimpleTBwGenTypeOption.scala:401)
```

Is IMO less descriptive than require saying TB is incorrect. Now I actually have to see what line 401 is... Granted, that check is specifically for TB generation since I use almost the same args elsewhere... but in case the Expect fails somewhere, I can't differentiate. Anyways, I don't really care that much either way. Will still push changes...

grebe: Sure thing, it was a nitpick. Did you look at putting it into src/test/resources? It would make it easier to e.g. make sure a test case works with some tool

grebe:
Great! Nice job, merge whenever you're ready

shunshou:
Yep did that. Apparently scala only makes it easy to get stuff from
resources in 2.12, not anything before, but w/e.

grebe:
Scala is weird about things

PR #67, opened 2017-01-29
Add ops

shunshou:
[Do NOT merge]


shunshou:
@grebe plz check last commit to double check I'm not doing something stupid.

src/main/scala/dsptools/numbers/algebra_types/Signed.scala
```diff
@@ -29,23 +29,21 @@ trait Signed[A] extends Any {
     val s = signum(a)
     !s.eq && !s.lt
   }
-  def isSignNegative(a: A): Bool = {
-    val s = signum(a)
-    !s.eq && s.lt
-  }
+  def isSignNegative(a: A): Bool = signum(a).lt
```
shunshou: @grebe double check changes here! Negative doesn't need !s.eq. Also, isSignNonPositive was originally incorrect (should be lt or eq)

grebe: Your changes look correct to me. It should be the case that `!s.eq`, right? Maybe there should be assertions.

src/main/scala/dsptools/numbers/UIntTypeclasses.scala
```diff
@@ -3,99 +3,116 @@
...
+  override def isSignNegative(a: UInt): Bool = false.B
+  // isSignNonZero, isSignNonPositive, isSignNonNegative derived from above (!)
 }
+
 trait UIntIsReal extends Any with IsIntegral[UInt] with UIntOrder with UIntSigned with hasContext {
   def toDouble(a: UInt): DspReal = ???
+  // In IsIntegral: ceil, floor, round, truncate (from IsReal) already defined as itself; 
+  // isWhole always true
+  
+  def isOdd(a: UInt): Bool = a(0)
```
grebe: Just thought of this now... what happens if there is a zero-width wire?

src/main/scala/dsptools/numbers/UIntTypeclasses.scala
```diff
@@ -3,99 +3,116 @@
...
+  // isEven derived from isOdd
+  // Note: whatever Chisel does -- is there overflow?
```
grebe: What do you mean by overflow in this context?

src/main/scala/dsptools/numbers/UIntTypeclasses.scala
```diff
@@ -3,99 +3,116 @@
...
+  // Generally better to use your own mod if you know input bounds
```
grebe: Should this be done in chisel or firrtl? It would be cool if you used the range API and a firrtl pass replaced a mod with a Mux(test, a, a - b) or whatever

src/main/scala/dsptools/numbers/UIntTypeclasses.scala
```diff
@@ -3,99 +3,116 @@
...
 trait ConvertableToUInt extends ConvertableTo[UInt] with hasContext {
   def fromShort(n: Short): UInt = n.toInt.U
-  //def fromAlgebraic(n: Algebraic): UInt = n.toBigInt.U
   def fromBigInt(n: BigInt): UInt = n.U
   def fromByte(n: Byte): UInt = n.toInt.U
-  def fromDouble(n: Double): UInt = n.toInt.U
-//  def fromReal(n: Real): UInt = n.toInt.U
-  //def fromRational(n: Rational): UInt = n.toInt.U
-  def fromType[B](n: B)(implicit c: ConvertableFrom[B]): UInt = c.toBigInt(n).U
+  def fromDouble(n: Double): UInt = n.round.toInt.U
```
grebe: Is it better to round here? Do you have a specific case in mind where it is bad not to round?

src/main/scala/dsptools/numbers/UIntTypeclasses.scala
```diff
@@ -3,99 +3,116 @@
...
   def fromInt(n: Int): UInt = n.U
-  def fromFloat(n: Float): UInt = n.toInt.U
-  def fromBigDecimal(n: BigDecimal): UInt = n.toBigInt.U
+  def fromFloat(n: Float): UInt = n.round.toInt.U
+  def fromBigDecimal(n: BigDecimal): UInt = n.doubleValue.round.toInt.U
   def fromLong(n: Long): UInt = n.U
+  def fromType[B](n: B)(implicit c: ConvertableFrom[B]): UInt = c.toBigInt(n).U
+  override def fromDouble(d: Double, a: UInt): UInt = {
```
grebe: Why is there a two argument version of this? It seems `a` is ignored. Seems like it should call `fromDouble(d)` (and the `require` could be moved there)

src/main/scala/dsptools/numbers/UIntTypeclasses.scala
```diff
@@ -3,99 +3,116 @@
...
+    require(d >= 0, "Double literal to UInt needs to be >= 0")
+    d.round.toInt.U
+  }
+  override def fromDoubleWithFixedWidth(d: Double, a: UInt): UInt = {
```
grebe: Is there a good reason for `a` to be a `UInt` rather than a width? If the width is unknown it will crash here rather than in user code.

src/main/scala/dsptools/numbers/UIntTypeclasses.scala
```diff
@@ -3,99 +3,116 @@
...
+    require(d >= 0, "Double literal to UInt needs to be >= 0")
+    val intVal = d.round.toInt  
+    val intBits = BigInt(intVal).bitLength
+    require(intBits <= a.getWidth, "Lit can't fit in prototype UInt bitwidth")
```
grebe: We talked about adding `require` for getWidth

src/main/scala/dsptools/numbers/implicits/AllOps.scala
```diff
@@ -7,15 +7,19 @@ import spire.macros.Ops
...
   * Java Bools. I suppose a more general solution would be generic in
   * return type, but the use cases there seem obscure.
   */
 
-final class EqOps[A <: Data](lhs:A)(implicit ev:Eq[A]) {
-  def ===(rhs:A): Bool = macro Ops.binop[A, Bool]
-  def =!=(rhs:A): Bool = macro Ops.binop[A, Bool]
+final class EqOps[A <: Data](lhs: A)(implicit ev: Eq[A]) {
+  def ===(rhs: A): Bool = macro Ops.binop[A, Bool]
+  def =!=(rhs: A): Bool = macro Ops.binop[A, Bool]
```
grebe: I agree this was goofy, I'm pretty sure this came from spire

src/main/scala/dsptools/numbers/UIntTypeclasses.scala
grebe: We talked about this- there is a good reason

shunshou:
@grebe mind checking some more? I'm mostly done with uint, sint, fixed type classes... still need to add basic trim op = floor to some binary point for multiply and div2 (via dspcontext). If this seems ok, I'll prob try to merge master into this branch...

src/main/scala/dsptools/numbers/chisel_types/SIntTypeClass.scala
```diff
@@ -0,0 +1,141 @@
...
+  }
+  override def minus(f: SInt, g: SInt): SInt = {
+    val diff = {
+      if(context.overflowType == Grow) f -& g
+      else if (context.overflowType == Wrap) f -% g
+      else throw DspException("Saturating subtractor hasn't been implemented")
+    }
+    ShiftRegister(diff, context.numAddPipes)
+  }
+  def negate(f: SInt): SInt = minus(0.S, f)
```
grebe: I think SInt has unary negation, right? Should `negate` inherit the wrap/grow/saturate behavior of `minus`?

src/main/scala/dsptools/numbers/chisel_types/SIntTypeClass.scala
```diff
@@ -0,0 +1,141 @@
...
+/**
+  * Defines basic math functions for SInt
+  */
+trait SIntRing extends Any with Ring[SInt] with hasContext {
+  def zero: SInt = 0.S
+  def one: SInt = 1.S
+  def plus(f: SInt, g: SInt): SInt = {
+    // TODO: Saturating mux should be outside of ShiftRegister
+    val sum = {
+      if(context.overflowType == Grow) f +& g
```
grebe: Saw this in the previous round of reviews and didn't think of it, but this should probably be a match statement instead of if else if ... else

src/main/scala/dsptools/numbers/chisel_types/SIntTypeClass.scala
```diff
@@ -0,0 +1,141 @@
...
+  def fromFloat(n: Float): SInt = fromDouble(n.toDouble)
+  def fromBigDecimal(n: BigDecimal): SInt = fromDouble(n.doubleValue)
+  def fromLong(n: Long): SInt = fromBigInt(BigInt(n))
+  def fromType[B](n: B)(implicit c: ConvertableFrom[B]): SInt = fromBigInt(c.toBigInt(n))
+  def fromBigInt(n: BigInt): SInt = n.S
+  def fromDouble(n: Double): SInt = n.round.toInt.S  
+  // Second argument needed for fixed pt binary point (unused here)
+  override def fromDouble(d: Double, a: SInt): SInt = fromDouble(d)
+  override def fromDoubleWithFixedWidth(d: Double, a: SInt): SInt = {
+    require(a.widthKnown, "SInt width not known!")
```
grebe: :)

src/main/scala/dsptools/numbers/chisel_base_type_extension/ChiselBaseNum.scala
```diff
@@ -9,7 +9,13 @@ object ChiselBaseNum {
 trait ChiselBaseNum[A <: Data] extends Any {
```
grebe: Not sure about this name. This trait is really about power of two operations- maybe we should call it `BinaryRepresentation` or something that suggest that more.

shunshou:
screwed up typing. merged in master...

shunshou:
@grebe Everything in numbers *except* Complex + ComplexType should be more or less done. Can you take a look? Complex will be done hopefully by tomorrow morning; and then I'm going to refactor DspContext a bit (just to make it neater) and also prob replace DspTester w/ VerboseDspTester (getting rid of PeekAs since you guys said it was in your other repo).

shunshou:
Changes to be committed to DspComplex (affects people's DSP blocks):
- imaginary -> imag
- multiplyByJ(x) -> x.j
- divideByJ(x) -> x.conj
-DSPComplex.wire(real, imag) made private [i.e. only used by ops] -> Should be explicit in your own code via Wire(DspComplex(gen)) to generate wire where real and imag have gen types

@grebe @stevo

src/main/scala/dsptools/numbers/chisel_concrete/DspComplex.scala
```diff
@@ -0,0 +1,82 @@
...
+  def protoWithFixedWidth[T <: Data:Ring:ConvertableTo](c: Complex, t: T): DspComplex[T] = {
+    DspComplex(ConvertableTo[T].fromDoubleWithFixedWidth(c.real, t), 
+      ConvertableTo[T].fromDoubleWithFixedWidth(c.imag, t))
+  }
+
+}
+
+class DspComplex[T <: Data:Ring](val real: T, val imag: T) extends Bundle {
+  // Multiply by j
+  def j: DspComplex[T] = DspComplex.wire(-imag, real)
```
grebe: I think it is confusing to call this `j`, especially because the companion object also has a function called `j` which has different behavior. I'm not sure what to call it. `multiplyByJ` is verbose but impossible to misunderstand. Something like `x_j` or `multj` that is shorter might be OK too.

src/main/scala/dsptools/numbers/chisel_concrete/DspComplex.scala
```diff
@@ -0,0 +1,82 @@
...
+  // Divide by j
+  def divj: DspComplex[T] = DspComplex.wire(imag, -real)
```
grebe: I am concerned that reflection is going to see these things as elements in the bundle (in fact, couldn't it do this recursively and stack overflow?). This is why we had the dummy stuff. We should probably make this a `Record` now- is Jack the right person to ask about that?

shunshou:
last commit includes fixes per @grebe

shunshou:
Also, @chick, @grebe https://github.com/ucb-bar/dsptools/blob/add_ops/src/main/scala/dsptools/numbers/README.md was kind of what I was hoping for ages back. 1) Make people aware that ops exist without looking through the code and 2) simple example for all relevant ops. Ideally, I'd like to pull in some Chisel3 specific stuff (the last TODO chunk) just so people don't have to look through Chisel3 documentation to figure out how to do something in a Dsp Module, but I don't think I have the time to prioritize that. It'd be great if someone else could help populate with use cases when I'm done w/ this PR...

shunshou:
@stevo about to change Utilities to DspTesterUtilities. Also, I changed toBigIntUnsigned to:

```
// Converts negative values to their 2's complement version
  def toBigIntUnsigned(x: Double, totalWidth: Int, fractionalWidth: Int): BigInt = {
    val bi = FixedPoint.toBigInt(x, fractionalWidth)
    val neg = bi < 0
    val neededWidth = if (neg) bi.bitLength + 1 else bi.bitLength
    require(neededWidth <= totalWidth, "Double -> BigInt width larger than total width allocated!")
    if (neg) (1 << neededWidth) - bi 
    else bi
  }
```

I believe this is right, and idk if you still had the 32 bit issue, but this should've fixed that. 

Also changed how the other functions in the new DspTesterUtilities work. In general, it's very bad to use math.pow(2,n) for stuff. For BigInts, to get stuff right, you should really be doing BigInt(1 << n). At least that's from my past experience. 

Haven't tested my changes, and I might've screwed up the logic, but the principal is there. Ideally, you and @grebe should double check.

shunshou:
Ok, @grebe I'm calling it quits for tonight. What's left:
Overhaul DspTester in favor of VerboseDspTester functionality (aside from removing doPokeAs/doPeekAs, which should go in your block tester if it hasn't already, I assume all of the current tests won't break...)
Cleanup Driver code (mostly weird formatting?)
Go through BBFloat.v b/c I made it a bit messy when trying stuff. 
Finish up Dsp related docs (other stuff will have to wait)
Add more tests (although, probably less thorough than I originally wanted b/c I've already individually tested some ops w/ a sloppy TB and the Specs in existence seem to do an ok job...)

Note: If anything causes the behavior of your FFT/PFB to change, it'll be the fact that I changed the defaults to DspContext. I think they're much more reasonable than what was originally there, but you should check... esp. now that DspContext is actually used in the typeclasses...

src/main/scala/dsptools/DspTester.scala
```diff
@@ -3,259 +3,366 @@
...
-            val bigInt = peek(u)
-            Left(bigIntBitsToDouble(bigInt))
-          case r: FixedPoint =>
-            val bigInt = peek(u)
-            Left(toDoubleFromUnsigned(bigInt, r.getWidth, r.binaryPoint.get))
-          // TODO:
-          //case s: SInt =>
-          //  Left(peek(s).toDouble)
-          case _ =>
-            throw DspException(s"peek($data): data has unknown type ${data.getClass.getName}")
```
stevobailey: Is this functionality moved somewhere else, or is expected to live outside dsptools?

shunshou: I think a peek as UInt should exist elsewhere... (esp. if it's singularly for testing your DspBlock stuff, then it should go in your DspBlockTester) but you can debate with me on why you might not want that.

stevobailey: Yeah just saw your comment from a week ago mentioning this. That's fine, it's probably specific to craft (we use it to poke/peek AXI interfaces) and should be in rocket-dsp-utils.

shunshou:
Yea please search all comments w/ @stevo.

shunshou:
And of course, I awkwardly reference random stevo again. ~_~ my bad.

shunshou:
more thorough tests + readme aside, should be ready to merge... will discuss more on monday, but i hope @chick won't mind helping to write a more thorough tb. i have a list of edge cases that i want tested and ops that i want tried (with various dspcontexts)...

shunshou:
@hqjenny @chick this should be more or less done. I wrote up standard operators etc. that should work and be best-practice approach at the README here:

https://github.com/ucb-bar/dsptools/tree/add_ops/src/main/scala/dsptools/numbers

The README should be used to guide tests. 

Basically, the new DspTester should be used to test all of the supported operations for Data:RealBits [=UInt, SInt, FixedPoint, DspReal]. Some of these operations have functionality that depend on DspContext settings. In those cases, it should be checked that operations function properly for all overflow types, trim types, etc. supported. -- esp. with edge cases. 

I have some more details regarding edge cases, but that should provide some starting point.

chick:
@shunshou looks good so far, I'll do more tonight and tomorrow will focus on this.  Thanks!

PR #71, opened 2017-02-17
Make DspComplex a Record.

grebe:
Also, fix tests that were failing quietly.

@shunshou I'm working on getting everything building against the add_ops branch. So far not too many problems, but we had several instances of `imaginary` being called without () which is a compilation error as long as `dummy` is being used.

I took it as an opportunity to try the new `Record` thing out and it seems to be pretty simple. Unfortunately, some tests are failing, but they don't seem to be related to Record. Do these tests fail for you, or is it somehow b/c of my changes?

Some tests were failing quietly so I added should be (true) in some places.

Also note that I have a small PR against chisel-testers to get Record working with PeekPokeTester

Also, I haven't done so here, but I feel like I should deprecate `imaginary` to remind us to move everything over to `imag`. What do you think?

### DIFF
--- a/src/main/scala/dsptools/numbers/chisel_concrete/DspComplex.scala
+++ b/src/main/scala/dsptools/numbers/chisel_concrete/DspComplex.scala
@@ -7,6 +7,7 @@ import chisel3.experimental.FixedPoint
 import dsptools.DspException
 import implicits._
 import breeze.math.Complex
+import scala.collection.immutable.ListMap
 
 object DspComplex {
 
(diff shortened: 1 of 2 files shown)


shunshou:
Please also double check sbt-test with add_ops w/o your changes to verify that you're seeing the same thing that I am.

shunshou:
I ran your branch. A bunch of stuff failed:

[error] Failed tests:
[error] 	dsptools.numbers.ComplexOpSpecification
[error] 	SimpleTB.SimpleTBSpec
[error] 	dsptools.numbers.TypeclassSpec
[error] 	examples.SimpleComplexAdderSpec

They didn't fail in my add_ops, so something doesn't play nicely with Record.

shunshou:
IMO Just keep it as Bundle and just swap imaginary -> imag in your code if you are ok with it being deprecated anyways. I like your Lit check and good for adding should be (true) for the TypeclassSpec.

chick:
I thought other changes had obviated the dummy stuff /ucb-bar/chisel3#342
If that's the only reason you are making it a record I don't think it's necessary

shunshou:
Yeah I'm pretty sure I actually tested some stuff w/o dummy when I first went about modifying Complex, but in case it was still an issue, per Paul's suggestion, I added it back in.

grebe:
OK, good point. I'll close this.

The second part of the PR is still failing for me- I'll make a separate PR.

PR #72, opened 2017-02-17
TypeclassSpec tests were failing quietly

grebe:
Integer[T] and Signed[T] are failing for me


shunshou:
I'll take a look. Would be useful to still add in your Lit checks tho!

shunshou:
Talked to Paul. The tests that were failing were wrong to begin with. Need less copy pasta.

shunshou:
Still plz add in Lit check. Remove the failing tests or correct them. And fine, go ahead and add ConvertableTo with DspComplex. I still disagree that you should have Double -> DspComplex (with only Real nonzero) in favor of more explicit adds, subtracts, and multiplies with real/imag #'s but w/e. As long as it's cleanly integrated...

shunshou:
^ Unless these tests were updated, this shouldn't be merged. The tests themselves were wrong.

grebe:
Yeah, don't merge. I'll get to it eventually...

PR #74, opened 2017-02-18
Add ops chick

chick:
FixedPoint shift tests implemented
A few doc things fixed

### DIFF
--- a/src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
+++ b/src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
@@ -1,3 +1,5 @@
+// See LICENSE for license details.
+
 package SimpleTB
 
 import chisel3._
@@ -352,7 +354,7 @@ class SimpleTBSpec extends FlatSpec with Matchers {
     val opt = new DspTesterOptionsManager {
       dspTesterOptions = optionsPass.dspTesterOptions
       testerOptions = optionsPass.testerOptions
-      commonOptions = optionsPass.commonOptions.copy(targetDirName = "test_run_dir/lit_sint")
+//      commonOptions = optionsPass.commonOptions.copy(targetDirName = "test_run_dir/lit_sint")
     }
     dsptools.Driver.execute(() => new SimpleLitModule(p.genShortS, p.genLongS, includeR = true, p), opt) { c =>
       new PassLitTester(c)
@@ -364,7 +366,7 @@ class SimpleTBSpec extends FlatSpec with Matchers {
     val opt = new DspTesterOptionsManager {
       dspTesterOptions = optionsPass.dspTesterOptions
       testerOptions = optionsPass.testerOptions
-      commonOptions = optionsPass.commonOptions.copy(targetDirName = "test_run_dir/lit_fix")
+//      commonOptions = optionsPass.commonOptions.copy(targetDirName = "test_run_dir/lit_fix")
     }  
     dsptools.Driver.execute(() => new SimpleLitModule(p.genShortF, p.genLongF, includeR = true, p), opt) { c =>
       new PassLitTester(c)
@@ -376,7 +378,7 @@ class SimpleTBSpec extends FlatSpec with Matchers {
     val opt = new DspTesterOptionsManager {
       dspTesterOptions = optionsFail.dspTesterOptions
       testerOptions = optionsFail.testerOptions
-      commonOptions = optionsFail.commonOptions.copy(targetDirName = "test_run_dir/lit_fix")
+//      commonOptions = optionsFail.commonOptions.copy(targetDirName = "test_run_dir/lit_fix")
     }
     dsptools.Driver.execute(() => new SimpleLitModule(p.genShortF, p.genLongF, includeR = true, p), opt) { c =>
       new FailLitTester(c)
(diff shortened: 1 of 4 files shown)


src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
```diff
@@ -349,7 +351,7 @@ class SimpleTBSpec extends FlatSpec with Matchers {
     val opt = new DspTesterOptionsManager {
       dspTesterOptions = optionsPass.dspTesterOptions
       testerOptions = optionsPass.testerOptions
-      commonOptions = optionsPass.commonOptions.copy(targetDirName = "test_run_dir/lit_sint")
+//      commonOptions = optionsPass.commonOptions.copy(targetDirName = "test_run_dir/lit_sint")
```
shunshou: Do these work without the directories now?

src/test/scala/dsptools/numbers/FixedPointSpec.scala
```diff
@@ -26,26 +28,109 @@ class FixedRing1(val width: Int, val binaryPoint: Int) extends Module {
...
-  println(s"Increment is $increment")
   for(i <- -2.0 to 3.0 by increment) {
-    println(s"Testing value $i")
-
     poke(c.io.in, i)
 
     expect(c.io.floor, breeze.numerics.floor(i), s"floor of $i should be ${breeze.numerics.floor(i)}")
     expect(c.io.ceil, breeze.numerics.ceil(i), s"ceil of $i should be ${breeze.numerics.ceil(i)}")
     expect(c.io.isWhole, breeze.numerics.floor(i) == i , s"isWhole of $i should be ${breeze.numerics.floor(i) == i}")
     expect(c.io.round, breeze.numerics.round(i), s"round of $i should be ${breeze.numerics.round(i)}")
```
shunshou: Should also double check behavior for 0.5, 1.5, etc. -- I'm not sure what Breeze's round is. My round is round half up. The behavior might be different if Breeze's round is round half to larger absolute value or there are also round half even/round half odd.

src/test/scala/dsptools/numbers/FixedPointSpec.scala
```diff
@@ -26,26 +28,109 @@ class FixedRing1(val width: Int, val binaryPoint: Int) extends Module {
...
+        }
+      }
+    }
+
+    "The shift family" - {
+      for (binaryPoint <- 0 to 16 by 2) {
+        s"should work with binary point $binaryPoint, for positives even when binaryPoint > width" in {
+          dsptools.Driver.execute(
+            () => new FixedPointShifter(8, binaryPoint = binaryPoint, fixedShiftSize = 2),
+            Array("-tdb", "2")
```
shunshou: Also double check with Verilator -- that gets things down the full pipe. FirrtlInterpreter wouldn't catch any translation issues from low Firrtl to Verilog, right?

src/test/scala/dsptools/numbers/FixedPointSpec.scala
```diff
@@ -26,26 +28,109 @@ class FixedRing1(val width: Int, val binaryPoint: Int) extends Module {
...
+          } should be(true)
+        }
+      }
+    }
+
+    "The shift family" - {
+      for (binaryPoint <- 0 to 16 by 2) {
+        s"should work with binary point $binaryPoint, for positives even when binaryPoint > width" in {
+          dsptools.Driver.execute(
+            () => new FixedPointShifter(8, binaryPoint = binaryPoint, fixedShiftSize = 2),
```
shunshou: In addition to sweeping the input value + BP, you should also be sweeping fixedShiftSize across tests from 0 to > width of input.

src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
chick: I don't think so, I am working on a fix in chisel-testers, but it was taking to much time so I've jumped to this for now.  But I will get back to it.

src/test/scala/dsptools/numbers/FixedPointSpec.scala
chick: @shunshou I am not sure what you mean here, I am incrementing by the smallest increment possible for the binaryPoint under consideration

src/test/scala/dsptools/numbers/FixedPointSpec.scala
chick: shift right firrtl spec says shift must be strictly less than, width of target.  Makes constructing this test a little more complicated if it's going to sweep for larger values of other shifts.  I'll see what I can do.

src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
shunshou: Ok

src/test/scala/dsptools/numbers/FixedPointSpec.scala
shunshou: Ok, as long as round is tested for numbers like 1.5, 2.5, -1.5, -2.5, etc. I expect it to always round half up i.e 1.5 -> 2, -1.5 -> -1. This is different from convention where -1.5 -> -2 (at least that we learn in grade school? Actually I'm not sure what we learned for rounding negatives...). There are other "round" methods like round half even so 1.5 -> 2, 2.5 -> 2, etc. It just needs to be readily apparent what we do "round half up" is fully descriptive vs. what Breeze does (I'm not sure what Scala math.round does).

src/test/scala/dsptools/numbers/FixedPointSpec.scala
shunshou: If that's what the Firrtl spec says, that's fine -- I just hope it has a compile time error then. And probably good to state somewhere in the readme.

PR #76, opened 2017-03-02
Shift tests, some DspContexts tests. UInt minus grow exception

chick:
I'd like to add my tests as a series of smaller PR's 
So this is the first, of more to come


chick:
Arggh, my merge of master blew up the commit list.  I tried rebasing but that did not go so well.  I think it's ok and will be squashed when merged

src/main/scala/dsptools/numbers/chisel_types/UIntTypeClass.scala
```diff
@@ -26,11 +26,11 @@ trait UIntRing extends Any with Ring[UInt] with hasContext {
   }
   override def minus(f: UInt, g: UInt): UInt = {
     val diff = context.overflowType match {
-      case Grow => f -& g
-      case Wrap => f -% g
+      case Grow => throw DspException("OverflowType Grow is not supported for UInt subtraction")
+      case Wrap => f.asSInt -% g.asSInt
```
shunshou: Do you need to first cast as SInt? Doesn't that make 4 bit inputs result in 6 bit outputs?

chick: Hmm, the test case of 1 - 2 where 1 and 2 are UInt<4> returns 15 with overflow wrap, which I believe means it's being kept to4 bits.  But I also just tested line 30 as ```f -% g``` and I still got 15.

chick: @shunshou I guess I will got with the ```f -% g``` as it seems more direct and seems to work in all my tests

shunshou:
I'm confused -- wasn't the issue that Chisel's -% returned SInt? Or am I missing something?

chick:
@shunshou I just think you missing the context.  We are still using the Typeclass stuff wrapped around the UInt minus.  The UInt minus returns an SInt then the typeclass casts it back into a UInt.

shunshou:
Oh sorry -- I thought for some reason you got rid of the .asUInt at the end.

PR #77, opened 2017-03-03
Tests for absolute value (abs)

chick:
Check that it works for all values
Original negative value results when using abs and overflow type wrap
with SInt or FixedPoint at max negative value for the given width


shunshou:
If you want to be more specific, when it's the most negative number representable, abs will return itself for Wrap behavior.

PR #78, opened 2017-03-03
Tests for context nesting

chick:
Use overflow type setting at module and block level to show that override takes place properly
@shunshou 
 - Is this exhaustive enough
- I think that testing one kind of DspContext setting is enough to prove the nesting of the others, but do you agree

### DIFF
--- a/src/test/scala/dsptools/DspContextSpec.scala
+++ b/src/test/scala/dsptools/DspContextSpec.scala
@@ -2,7 +2,10 @@
 
 package dsptools
 
-import org.scalatest.{Matchers, FreeSpec}
+import chisel3._
+import dsptools.numbers.Ring
+import dsptools.numbers.implicits._
+import org.scalatest.{FreeSpec, Matchers}
 
 //scalastyle:off magic.number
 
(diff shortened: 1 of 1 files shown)


src/test/scala/dsptools/DspContextSpec.scala
```diff
@@ -49,4 +52,71 @@ class DspContextSpec extends FreeSpec with Matchers {
...
+    io.outWrap := io.in1 + io.in2
+  }
+  DspContext.withOverflowType(Grow) {
+    io.outGrow := io.in1 + io.in2
+  }
+
+  io.outDefault := io.in1 + io.in2
+}
+
+class ContextNestingTop[T <: Data : Ring](gen1: T, gen2: T) extends Module {
```
chick: @grebe @shunshou If I change this to ```[T <: UInt : Ring]``` the tests here fail because it doesn't end up using the typeclass stuff and thus does not implement the DspContext specified.  This seems dangerous and more than a little hard to describe.  Any ways we can make this more robust

shunshou: Yeah... you exposed a real problem with these type classes, but I have no idea what to do about it...

shunshou:
I think testing one DspContext setting is sufficient, but it should probably be done using funkier operations. I haven't really looked -- I assume you picked cases where a bad first Wrap would cause the output to fail when the output is Grow or something?

src/test/scala/dsptools/DspContextSpec.scala
shunshou: Is it possible to change the type classes to be on T <: ( U <: Data) : Ring or something like that?

chick:
@shunshou I focused these tests on wrap because it is easy to detect which result you got by adding a number to a uint at the high boundary.

PR #79, opened 2017-03-03
Adds link to numbers readme from top-level readme

chick:
This is for my convenience, but I think it is a good way to advertise that the numbers readme exists


shunshou:
Yeah no worries. You could also move it out.

PR #81, opened 2017-03-11
Adds some more tests on div2, mul2 and signBit and fixes compile problems

chick:
Fixed up apparent change in RegNext location
Fix numbers readme for signBit
Add some tests for div2, mul2 signBit
Cleaned up deprecation
Reg(next -> RegNext
log2Up -> log2Ceil
This commits causes a firrtl internal error on the ResetNSpec.  I have created issue there per the dump instructions


shunshou:
lgtm -- instead of log2ceil (or in addition to), it might be useful to try specifying ranges (since you're essentially converting ranges to bitlength anyways...).

shunshou:
I'm not sure which commit this came from -- but why are all of your Expect strings:

```
STEP 1x -> 194
dynamic shift left 3 of -107.0 should be ${expectedValue(value, left = true, dynamicShiftValue)}  EXPECT FixedPointShifter.io_dynamicShiftLeftResult -> -856.0 == E -856.0 PASS, tolerance = 0.0, Q14.0
dynamic shift right 3 of -107.0 shouldbe -14.0  EXPECT FixedPointShifter.io_dynamicShiftRightResult -> -14.0 == E -14.0 PASS, tolerance = 0.0, Q7.0
  
```

${expectedValue blah blah ?? Is that intentional?

PR #83, opened 2017-03-12
Temp resolution to Ring ops without using Gen

shunshou:
So that you still get DspContext. Must use context_+, etc. Maybe different names are better... This still has issues b/c you're not preventing people from using +, so they could still think + gets DspContext when it doesn't...


shunshou:
Nope and I don't plan to add any. It just points context_-, etc. to the appropriate Ring ops. If I run into bugs, I'll fix it then. Someone who better understands typeclasses should be able to judge if I screwed up.

I talked to Stephen in excess last night, and I think the conclusion is that Typeclasses are not meant to change functionality; only add onto. The + should be Chisel +, etc. 

The fact that there's confusion about how things works and what things do kind of proves that point.  I'm just adding these as a hack so I have even a 1% chance of meeting my tapeout deadline.

PR #84, opened 2017-03-15
Ring refactor

shunshou:
@chick Making this a PR. Looks like this also contains #81?


chick:
Yes, it appears to contain #81, I think that's ok, though I can't say I intended to start the branch there

chick:
@shunshou If fixed the numbers spec, but there are still some more tests failing here. I will get them asap, need a break right now

shunshou:
Yeah no worries. I'm more concerned about the chisel/testers stuff...

chick:
All tests run now except for ResetNSpec
@shunshou @grebe can you take a quick look at what I did for context_abs

src/main/scala/dsptools/numbers/algebra_types/Signed.scala
```diff
@@ -42,6 +43,11 @@ object Signed {
 
 private class OrderedRingIsSigned[A <: Data](implicit o: Order[A], r: Ring[A]) extends Signed[A] {
   def signum(a: A): ComparisonBundle = o.compare(a, r.zero)
-  def abs(a: A): A = Mux(signum(a).lt, r.negate(a), a)
+  def abs(a: A): A = {
+    Mux(signum(a).lt, r.negate(a), a)  //TODO: should this be negateContext
+  }
+  def context_abs(a: A): A = {
```
shunshou: You need to pipeline signum(a).lt, etc. the same as negateContext for UInt stuff

src/main/scala/dsptools/numbers/algebra_types/helpers/Sign.scala
```diff
@@ -69,6 +69,7 @@ object Sign {
     override def sign(a: Sign): Sign = a
     def signum(a: Sign): ComparisonBundle = ComparisonHelper(a.zero, a.neg)
     def abs(a: Sign): Sign = if (a == Negative) Positive else a
+    def context_abs(a: Sign): Sign = if (a == Negative) Positive else a
```
shunshou: Im not sure that's necessay... (also only included b/c Paul wanted Spire "Sign", but I can't think of a use case...)

src/main/scala/dsptools/numbers/chisel_types/DspRealTypeClass.scala
```diff
@@ -47,6 +47,7 @@ trait DspRealSigned extends Any with Signed[DspReal] with hasContext {
     ComparisonHelper(a === DspReal(0.0), a < DspReal(0.0))
   }
   def abs(a: DspReal): DspReal = Mux(a < DspReal(0.0), DspReal(0.0) - a, a)
+  def context_abs(a: DspReal): DspReal = Mux(a < DspReal(0.0), DspReal(0.0) - a, a)
```
shunshou: a < DspReal(0.0), a need to be delayed by context.numAddPipes (since DspReal(0.0) - 1 uses that number of add registers. Delayed via Shift. Otherwise there will be horrible timing mismatch.

src/main/scala/dsptools/numbers/chisel_types/FixedPointTypeClass.scala
```diff
@@ -186,6 +186,7 @@ trait FixedPointReal extends FixedPointRing with FixedPointIsReal with Convertab
   override def fromInt(n: Int): FixedPoint = super[ConvertableToFixedPoint].fromInt(n)
   // Overflow only on most negative
   def abs(a: FixedPoint): FixedPoint = Mux(isSignNegative(a), super[FixedPointRing].minus(zero, a), a)
+  def context_abs(a: FixedPoint): FixedPoint = Mux(isSignNegative(a), super[FixedPointRing].minusContext(zero, a), a)
```
shunshou: Same here. Need to delay all the other paths the same way.

src/main/scala/dsptools/numbers/chisel_types/SIntTypeClass.scala
```diff
@@ -67,6 +72,13 @@ trait SIntSigned extends Any with Signed[SInt] with hasContext {
   }
   // isSignPositive, isSignNonZero, isSignNonPositive, isSignNonNegative derived from above (!)
   // abs requires ring (for overflow) so overridden later
+  def context_abs(a: SInt): SInt = {
+    context.overflowType match {
+      case Grow => Mux(a >= 0.S, a, 0.S -& a)
```
shunshou: Do you want to use context + context_-? context_- obviously also includes a shift register...

src/main/scala/dsptools/numbers/chisel_types/SIntTypeClass.scala
```diff
@@ -130,7 +142,7 @@ trait SIntInteger extends SIntRing with SIntIsReal with ConvertableToSInt with
   // fromSInt also included in Ring
   override def fromInt(n: Int): SInt = super[ConvertableToSInt].fromInt(n)
   // Overflow only on most negative
-  def abs(a: SInt): SInt = Mux(isSignNegative(a), super[SIntRing].minus(0.S, a), a)
+  def abs(a: SInt): SInt = Mux(isSignNegative(a), super[SIntRing].minusContext(0.S, a), a)
```
shunshou: Why is abs using minusContext? 

abs should use the correct one; 

abs_context should use minusContext.

src/main/scala/dsptools/numbers/algebra_types/Signed.scala
grebe: For operations like + and -, don't you sometimes pipeline every three adds, or something like that? Perhaps we simply shouldn't have a definition here.

src/main/scala/dsptools/numbers/algebra_types/helpers/Sign.scala
grebe: As currently implemented not sure how useful it is. I think I left it in at the time wondering if it would be useful if you wanted to use signed-magnitude notation (e.g. ldpc). I won't be mad if sign just gets ripped out. Maybe that should be in a separate PR though

src/main/scala/dsptools/numbers/chisel_types/DspComplexTypeClass.scala
```diff
@@ -12,30 +12,51 @@ class DspComplexRing[T <: Data:Ring] extends Ring[DspComplex[T]] with hasContext
   def plus(f: DspComplex[T], g: DspComplex[T]): DspComplex[T] = {
     DspComplex.wire(f.real + g.real, f.imag + g.imag)
   }
+  def plusContext(f: DspComplex[T], g: DspComplex[T]): DspComplex[T] = {
+    DspComplex.wire(f.real context_+ g.real, f.imag context_+ g.imag)
+  }
   def times(f: DspComplex[T], g: DspComplex[T]): DspComplex[T] = {
+    val c_p_d = g.real + g.imag
```
grebe: This feels like unnecessary code duplication. I'm wondering if we should have a function `threeMulComplexMultiply` and `fourMulComplexMultiply` that also takes an addition and multiplication function as parameters.

Then the `times(f, g)` would call `threeMulComplexMultiply(f, g, plus, multiply)` and `timesContext` would look at the context and either call `threeMulComplexMultiply(f, g, contextPlus, contextMultiply)` or `fourMulComplexMultiply(f, g, contextPlus, contextMultiply)` (with appropriate extra shift registers). What do you think?

src/main/scala/dsptools/numbers/implicits/AllOps.scala
```diff
@@ -119,4 +120,11 @@ class BinaryRepresentationOps[A <: Data](lhs: A)(implicit ev: BinaryRepresentati
   def div2(n: Int): A = ev.div2(lhs, n)
   def mul2(n: Int): A = ev.mul2(lhs, n)
   def trimBinary(n: Int): A = ev.trimBinary(lhs, n)
+}
+
+class ContextualRingOps[A <: Data](lhs: A)(implicit ev: Ring[A]) {
```
grebe: This is a good name :)

src/main/scala/dsptools/numbers/algebra_types/Signed.scala
shunshou: Well, from our conversation last time, in that case, you'd set context.numAddPipes to 0 and handle externally... (at least I thought that's what we decided would be safest?)

src/main/scala/dsptools/numbers/chisel_types/DspComplexTypeClass.scala
shunshou: I'm ok with either. Just be sure the shift register stuff is consistent...

src/main/scala/dsptools/numbers/algebra_types/Signed.scala
grebe: Oh yeah, I forgot about that. You're right

shunshou:
There's a bug in DspTester. I'm not sure if this will screw things up, but I'm going to push a fix here, since I'm on this branch.

src/main/scala/dsptools/numbers/algebra_types/Signed.scala
chick: @shunshou Can you point me to an example what it means to do this?  I'm looking around but I don't quite get it

src/main/scala/dsptools/numbers/algebra_types/helpers/Sign.scala
chick: If I don't have context_abs here then I get a SignAlgebra must be abstract error.

src/main/scala/dsptools/numbers/algebra_types/Signed.scala
shunshou: Check out what I did with the 3 multiply version of complex multiply. I had to do something like (I forget the ops) a + c*d but since c *d is delayed by numMulPipes, I needed to delay a by the same amount. Basically you should use shiftregister(a, context.numAddPipes) instead of a. And likewise for the condition. You should be able to catch delay misalignments if you test contexts with delays. That's actually probably pretty important.

src/main/scala/dsptools/numbers/chisel_types/DspComplexTypeClass.scala
chick: I'm feeling weak on my complex arithmetic, @grebe can you template this out for me a little if it's important

src/main/scala/dsptools/numbers/chisel_types/DspRealTypeClass.scala
chick: @shunshou This shift should be done for both abs and context_abs right?

src/main/scala/dsptools/numbers/chisel_types/FixedPointTypeClass.scala
chick: @shunshou This shift should be done for both abs and context_abs right?

src/main/scala/dsptools/numbers/chisel_types/SIntTypeClass.scala
chick: @shunshou I wondered if I used context_- then I wouldn't need to do the match statement, because the context_- would do the right thing.  Is that right?  (with ShiftRegister, of course)

src/main/scala/dsptools/numbers/chisel_types/SIntTypeClass.scala
chick: @shunshou good catch, I think this is wrong, will fix.  I still have the same question as elsewhere about whether the straight abs needs the delays

src/main/scala/dsptools/numbers/algebra_types/Signed.scala
chick: @shunshou it sounds important.  Can you give me some idea of how I would test that.  Is this some sort of analytic firrtl pass

chick:
@shunshou I pushed a new commit to this PR, questions above still there, but you can now see what I've done so far

shunshou:
@chick did I answer all of your questions?

chick:
@shunshou I don't think so but allow me to summarize.

- [ ] Do both abs and contexts_abs get ShiftRegister applied to *a* side of Mux
- [ ] For SInt context_abs, do I need to bother with the match statement on context if I use context_- 
- [ ] How can I test that the ShiftRegisters are done correctly, you suggested this is an important test
- [ ] I am a little unclear as to what needs to happen for DspComplexRing *times*

shunshou:
No, only context_abs. If you use abs, you're using the default subtract, which doesn't rely on numAddPipes. context_abs should use context_sub, which relies on numAddPipes, so that means a < 0 and a must be ShiftRegister-ed. Check back on a previous comment. I gave a very specific implementation comment. 

For context_abs, you should just use context_- -- no need to use match. 

I also mentioned a test for ShiftRegisters being done correctly. i.e. try DspComplex context_* with different numAddPipes, numMulPipes. DspComplex consists of multiply then add -- so the total delay should be numAddPipes + numMulPipes. Make sure that for whatever numAddPipes + numMulPipes you use, after numAddPipes + numMulPipes delays, see that the answer is right. To sanity check, purposely drop ShiftRegister(a) on one side of the add and see that for numMulPipes > 0, the test should fail (assuming you're updating the input to a different value every cycle).

I think the implementation that currently exists is right. @grebe just has gripes about copy pasting. I'd ignore that for now if you're not sure what he means and just make a TODO cleanup thing that Paul can deal with after tapeout.

chick:
@shunshou @grebe 
Sorry to keep pushing back on this one but, I want to make sure I get it right
Currently in SIntTypeClass My version of context_abs looks like
```scala
  def context_abs(a: SInt): SInt = {
    def localNegate(g: SInt): SInt = {
      val diff = context.overflowType match {
        case Grow => 0.S -& g
        case Wrap => 0.S -% g
        case _ => throw DspException("Saturating abs hasn't been implemented")
      }
      ShiftRegister(diff, context.numAddPipes)
    }
    Mux(a >= 0.S, ShiftRegister(a, context.numAddPipes), localNegate(a))
  }
```
I could not figure out a way to coerce contextMinus in SIntRing to be available

shunshou:
Move context_abs down to something else (kind of like all my comments where I said x depends on y, so I don't do it here).

chick:
@shunshou I moved it and added a test ShiftRegisterDelaySpec that is a model for testing that both sides of the context_abs mux has the same delay.   Can you comment on it's overall design before I put more time into it

chick:
@shunshou This PR, builds and seems kind of ok, how much more do we need to do to it to feel ok about merging it and do additional work in a new branch/PR

chick:
@shunshou There are bunch of muxes in chisel_concrete.DspReal that do not have ShiftRegisters or context_<op>, I am assuming this is OK, because this stuff is not synthesizable.  Let me know if otherwise.

shunshou:
This is not ok, b/c timing behavior between DspReal and FixedPoint should be consistent.

chick:
@shunshou ok, DspReal should ShiftRegisters

chick:
@shunshou Next question.  Right now things like ceil and floor do not have context_<op> variants which means that a ceil on fixed point will call a minus (not minusContext) within a Mux that does not have the other terms wrapped in shift registers.  What should I do for this, create a parallel family to make sure a developer can choose to use with timing delays or not

shunshou:
IMO for ops that aren't in base Chisel, like ceil, floor, you should just change their behavior to use context_minus, etc, and have users expect that they take context info for delay, growth, etc. @grebe should chime in if he disagrees.

chick:
@shunshou @grebe Here is next question.  I'm testing that delays match between DspFixed and DspReal, DspReal has a ceil method in it ```DspRealIsReal``` trait, but that never gets called because the scala compiler just finds the ceil method in DspReal, this causes DspReal to not be delayed the same as FixedPoint. 
It seems like maybe I should create a ```context_ceil```.  I think this would work but then I'd have to back propagate context_ceil, through ```IsRealOps``` and make sure everybody else dispatches this correctly.  Any other ideas?

shunshou:
Change the name of DspReal ceil to DspReal_ceil_int or something and scope that it's only usable in the dsptools.numbers package?

shunshou:
@chick I hope I didn't screw this up. I made wire public again. Trying to keep up to date with this branch...

PR #101, opened 2017-09-07
Split BlackBoxFloat.v into a file for each module

### LINKED ISSUE #93: BlackBoxFloat Ln and Log10
VCS is super unhappy with them (not sure how you can constrain the input to be >0 b/c it keeps saying it's invalid for in < 0)... I've been commenting them out every time I run tests, which kind of sucks. I recall @ducky64 had a PR that pulled ops out (so you wouldn't be importing all of the Real SystemVerilog ops if you only used one or two). We should consider merging ASAP.

grebe:
Using blackbox implementations has been an all-or-nothing proposition. Some modules might throw errors if disconnected (e.g. log) and some waveform viewers will show disconnected blackboxes as top level IOs.

All of this is annoying.

Splitting the implementations up so each module lives in its own file avoids all of these problems.

This should close #93.


src/main/scala/dsptools/numbers/blackbox_compatibility/DspRealVerilatorBB.scala
```diff
@@ -36,85 +35,87 @@ class BlackboxTwoOperandBool extends BlackBoxWithVerilog {
...
-class BBFFromInt extends BlackBoxWithVerilog {
-  val io = IO(new Bundle() {
-    val in = Input(UInt(DspReal.underlyingWidth.W))
-    val out = Output(UInt(DspReal.underlyingWidth.W))
-  })
-  addVerilog()
-}
+class BBFFromInt extends BlackboxOneOperand
 
-class BBFToInt extends BlackBoxWithVerilog {
```
grebe: @chick, can you remember if there was a reason this wasn't BlackboxOneOperand?

src/test/scala/dsptools/numbers/BlackBoxFloat.scala
```diff
@@ -249,7 +249,11 @@ class FloatOpTester[T <: FloatOps](c: T, testTrigFuncs: Boolean = true) extends
 class BlackBoxFloatSpec extends ChiselFlatSpec {
   "A BlackBoxed FP block" should "work" in {
     assertTesterPasses({ new BlackBoxFloatTester },
-        Seq("/BlackBoxFloat.v"))
+        Seq(
+          "/BBFAdd.v",
```
grebe: This is not the best

shunshou:
I'm glad people (or you?) are working through the issues. :D... finally... :D...

PR #102, opened 2017-09-12
Importing dsptools.numbers._ now includes implicits by default

### LINKED ISSUE #22: Should we add implicits to dsptools.numbers package object?
I'm finding that 99% of the dsptools code I write requires the following:
```
import dsptools._
import dsptools.numbers._
import dsptools.numbers.implicits._
```

I think there is a use case for using dsptools without the numbers, but I'm not so sure it makes sense to use `dsptools.numbers` without the implicits.

grebe:
This exposed some interesting little weird things. I'll add some inline comments.

Closes #22


src/main/scala/dsptools/numbers/chisel_concrete/DspReal.scala
```diff
@@ -107,6 +107,14 @@ class DspReal(lit: Option[BigInt] = None) extends Bundle {
     oneOperandOperator(Module(new BBFCeil()))
   }
 
+  def round(): DspReal = (this + DspReal(0.5)).floor()
```
grebe: DspReal has some things that call the typeclass, especially truncate. That seemed bad. This may change the current behavior, though, because the implementation of truncate here doesn't have pipeline registers.

src/main/scala/dsptools/numbers/chisel_types/DspRealTypeClass.scala
```diff
@@ -71,7 +71,7 @@ trait DspRealIsReal extends Any with IsReal[DspReal] with DspRealOrder with DspR
   def isWhole(a: DspReal): Bool = a === round(a)
   // Round *half up* -- Different from System Verilog definition! (where half is rounded away from zero)
   // according to 5.7.2 (http://www.ece.uah.edu/~gaede/cpe526/2012%20System%20Verilog%20Language%20Reference%20Manual.pdf)
-  def round(a: DspReal): DspReal = floor(a + DspReal(0.5))
+  def round(a: DspReal): DspReal = a.round()
   def truncate(a: DspReal): DspReal = {
```
grebe: Should this be called `context_truncate`?

src/main/scala/dsptools/numbers/package.scala
```diff
@@ -0,0 +1,14 @@
+package dsptools
+
+package object numbers extends AllSyntax with AllImpl with spire.syntax.RingSyntax
```
grebe: I read somewhere that sometimes people will have a lowercase package (I think) without the implicit and an uppercase version of the package (I think) with the implicits. Is that something we should think about doing?

PR #103, opened 2017-09-18
Update from release

ucbjrl:
Add publishing parameters from release (so they're available for editing in master).
Bump Scala version to 2.11.11
Use "current" published SNAPSHOT versions of Chisel dependencies.


grebe:
lgtm

PR #107, opened 2017-10-04
Enable DspReal constants and initialize test hardware in preparation for the InvalidateAPI.

ucbjrl:
Use the internal arithmetic classes to enable DspReal literals.
Add an optional argument to DspReal class constructor (to avoid recursion) when using the above to create literals.

### DIFF
--- a/src/test/scala/dsptools/numbers/BlackBoxFloat.scala
+++ b/src/test/scala/dsptools/numbers/BlackBoxFloat.scala
@@ -12,7 +12,7 @@ import dsptools.DspTester
 
 class BlackBoxFloatTester extends BasicTester {
   val (cnt, _) = Counter(true.B, 10)
-  val accum = RegInit(Wire(DspReal(1.0)))
+  val accum = RegInit(DspReal(0))
 
   private val addOut = accum + DspReal(1.0)
   private val mulOut = addOut * DspReal(2.0)
@@ -105,6 +105,9 @@ class FloatOps extends Module {
     val out = Output(DspReal(1.0))
     val boolOut = Output(Bool())
   })
+
+  io.boolOut := false.B
+  io.out := DspReal(0)
 }
 
 class FloatOpsWithTrig extends FloatOps {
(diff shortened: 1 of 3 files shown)


src/test/scala/dsptools/numbers/BlackBoxFloat.scala
```diff
@@ -12,7 +12,7 @@ import dsptools.DspTester
 
 class BlackBoxFloatTester extends BasicTester {
   val (cnt, _) = Counter(true.B, 10)
-  val accum = RegInit(Wire(DspReal(1.0)))
+  val accum = RegInit(DspReal(0))
```
chick: Did this change because it wasn't working correctly before (was getting 0 and not 1)
or is there something else here.

ucbjrl: It wasn't working "as advertised":
  `RegInit(Wire(DspReal(1.0)))` is equivalent to `Reg(DspReal())`
That is, the value of the DspReal.node  is effectively lost and the Reg is initialized with an unconnected wire. The generated FIRRTL is:
```
    wire _T_13 : {node : UInt<64>} @[BlackBoxFloat.scala 15:27]
    _T_13 is invalid @[BlackBoxFloat.scala 15:27]
    reg accum : {node : UInt<64>}, clock with : (reset => (reset, _T_13)) @[BlackBoxFloat.scala 15:22]
```
It was getting set to 0 as a side-effect of the `_T_13 is invalid` (I believe).

After our fix, the generated FIRRTL is:
```
    BBFAdd.out is invalid
    BBFAdd.in2 is invalid
    BBFAdd.in1 is invalid
    BBFAdd.in1 <= UInt<64>("h00") @[DspReal.scala 26:21]
    BBFAdd.in2 <= UInt<64>("h00") @[DspReal.scala 27:21]
    wire _T_16 : {node : UInt<64>} @[DspReal.scala 28:19]
    _T_16.node <= BBFAdd.out @[DspReal.scala 29:14]
    reg accum : {node : UInt<64>}, clock with : (reset => (reset, _T_16)) @[BlackBoxFloat.scala 15:22]
```
I suspect there's a better way to do this, but this is sufficient for my purposes.

grebe:
lgtm

PR #113, opened 2017-10-12
Fix toBigIntUnsigned

stevobailey:
Maybe I'm misinterpreting the point of this function, but the conversion seems wrong. When I tried toBigIntUnsigned(-1.0, 8, 5), I got 32 instead of what I expected (224). So instead the proposed change performs a standard 2's complement conversion.

### DIFF
--- a/src/main/scala/dsptools/tester/DspTesterUtilities.scala
+++ b/src/main/scala/dsptools/tester/DspTesterUtilities.scala
@@ -18,7 +18,7 @@ object DspTesterUtilities {
   def toBigIntUnsigned(x: Double, totalWidth: Int, fractionalWidth: Int): BigInt = {
     val bi = FixedPoint.toBigInt(x, fractionalWidth)
     val neg = bi < 0
-    val neededWidth = if (neg) bi.bitLength + 1 else bi.bitLength
+    val neededWidth = bi.bitLength + 1
     require(neededWidth <= totalWidth, "Double -> BigInt width larger than total width allocated!")
     if (neg) (BigInt(1) << totalWidth) + bi
     else bi
--- a/src/main/scala/dsptools/tester/DspTesterUtilities.scala
+++ b/src/main/scala/dsptools/tester/DspTesterUtilities.scala
@@ -20,7 +20,7 @@ object DspTesterUtilities {
     val neg = bi < 0
     val neededWidth = if (neg) bi.bitLength + 1 else bi.bitLength
     require(neededWidth <= totalWidth, "Double -> BigInt width larger than total width allocated!")
-    if (neg) (((BigInt(1) << totalWidth) - 1) ^ -bi) + BigInt(1)
+    if (neg) (BigInt(1) << totalWidth) + bi
     else bi
   }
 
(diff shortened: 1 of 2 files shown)


shunshou:
Seeing as how I didn't use this code anywhere... It looks like I made a mistake and just never caught it... Unless I intended it to return the absolute value... o_o (which doesn't seem right)...

shunshou:
Oh I think I get it. Basically, the minimum # of bits required to represent -1 with 5 fractional bits leads to 32 (and the output is treated as a 6-bit thing). When you do a .bitLength on the result, it should say 6. I just ignored the totalWidth arg, except maybe intended to use that as a check? If you replace neededWidth with totalWidth, I believe you should get the same result as your direct two's complement thing? [I can't math so late at night...; in general, that's how Chisel has taken care of signed to unsigned conversions...]. Kind of misleading... I'm sure at the time I thought I had a reason for doing that or was just being semantically inaccurate...

shunshou:
I agree with Chick. Would be good to add unit test for sanity check, then feel free to merge?

stevobailey:
Yeah I tried cloning and running sbt test but many of them failed. Is there
something special I need for the existing tests to work outright? I'll try
adding a test and using testOnly so we can merge in the PR.

On Oct 13, 2017 10:10 PM, "Angie Wang" <notifications@github.com> wrote:

I agree with Chick. Would be good to add unit test for sanity check, then
feel free to merge?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ucb-bar/dsptools/pull/113#issuecomment-336610805>, or mute
the thread

shunshou:
There's been too much stuff going on w/ chisel/firrtl, etc. If you don't have compatible versions, things will be very sad... But this doesn't require any Chisel-specific Scala, so yeah an sbt test-only unit test would be great!

grebe:
@stevobailey my dsptools passed tests with whatever i have publishLocal'd, fwiw

stevobailey:
Okay I added a test (and the existing tests work when I add verilator to my path haha). While writing the test, I decided that the width checker was incorrect. Before, it ignored the sign bit if the input was positive, so `toBigIntUnsigned(2, 4, 2)` and `toBigIntUnsigned(-2, 4, 2)` both gave `8` as the result, though `2` shouldn't fit into a 4.2 signed fixed point type. Thoughts?

stevobailey:
Also I verified that the previous code failed the test I wrote.

shunshou:
Hm... toBigIntUnsigned(2, 4, 2) = 1000. But it is unsigned... So if you check if this result is > 0, it is....  toBigIntUnsigned(-2, 4, 2) is also 1000..., as "unsigned" I guess "BigIntUnsigned" just means raw bits, and it's up to you to interpret it correctly...

I'm not really sure what the use case I was going for. I guess you can append a 0 for it to be unsigned... 

You can change the function name to be something more obvious? In a way that matches the actual thing being performed (append 0 if you want)...

grebe:
I feel like `toBigIntUnsigned` should only be used if the input is treated as signed, or else the conversion should be called `toBigInt`. `signedToBigIntUnsigned` might make that more clear, though. I don't have a strong preference either way, lgtm.

shunshou:
Yeah maybe rename it to signedToBigIntUnsigned and just make sure you have a 0 in front for positive #'s...

stevobailey:
Well the check for neededWidth vs. totalWidth ensures that a signed positive input will have a 0 in front if there are enough bits. I'll change the name and mergy merge.

stevobailey:
Also, for what it's worth, Travis passes with broken code...

PR #117, opened 2017-11-06
Remove dependency on plotly.

grebe:
@ucbjrl I thought there were references to plotly left over in the repo but it looks like they're all gone. Simple fix.


ucbjrl:
Passes my tests.

PR #126, opened 2018-05-02
Less verbose

grebe:
Remove output from some tests.

Now the output should be small enough that travis is happy.


grebe:
@shunshou is this OK?

PR #128, opened 2018-05-03
Add rocket-dsp-utils, switch to Mill

grebe:
Rocket-dsp-utils is currently a separate repo for dsp-type things that depend on rocket. I don't really like maintaining a separate dependency, so I've been working on making dsptools have two subprojects in one repo:
1) dsptools as it currently exists, without any rocket dependency
2) rocket-dsp-utils

This PR looks enormous because I'm merging the entire history for rocket-dsp-utils, but the only real changes are at the end.

This PR uses Mill, Li Haoyi's new Scala build tool. It already seems to be a big improvement on SBT in some ways that I care about. @ucbjrl, can you take a look at it? I'd like to talk with you about how it works with your flow for publishing.


shunshou:
I’m really not a fan of including rocket stuff :/ didn’t look at your
commit, but how is the Chisel utils refactor coming along?

grebe:
So rocket should now be usable as an sbt dependency (my PRs are merged/approved), so I think for the purposes of this repo the old pain points should be better. I think in the not-so-distant future rocket will be on sonatype. I don't forsee anything in here that actually plops down a rocket core and tries to run code on it, which would require duplicating the rocket build system.

I've been out of the loop on the chisel utils refactor- do you think stuff currently in dsptools should migrate in to chisel utils?

shunshou:
No. Should stay here. But stuff like Async FIFOs were going to be moved out
eventually (from rocket into chisel utils?)... Thought that was my
understanding.

grebe:
I think that's true of the FIFO, but I haven't heard that for memory interfaces, diplomacy, registerrouter, crossbars, and that kind of thing. That's the bigger dependency here.

grebe:
I'm coming back to this- I'm going to update this PR so it uses sbt and mill in parallel (like @ucbjrl has been doing with other repos here), but make it so that normal `sbt compile` or `sbt run` only compiles or runs the non-rocket stuff. To use rocket things, you'll need to explicitly choose to do so.

We need to keep rocket and non-rocket things totally separate as far as sbt is concerned. I think it would make life a lot better if all the code lived in one git repo, though, so I'd really like to get something like this PR merged in.

Any thoughts?

shunshou:
Guess that’s fine... still would rather pull utils out of rocket chip...
but who knows when/if that will ever happen...

grebe:
Defer to #131

PR #131, opened 2018-09-05
Add rocket dsp utils

grebe:
Adds rocket-dsp-utils as a subdirectory. `dsptools` is the default build target- compiling, testing, etc. don't pull any rocket things unless you explicitly choose the rocket package.

Subsumes #128 with the same idea, but sticking with SBT and deferring mill stuff for later.


shunshou:
Just to keep note: Ideally, rocket-dsp-utils and ACED will become separate
sub-projects under dsptools.

PR #136, opened 2018-09-25
Replace deprecated isLit

grebe:
Removes compiler warning


grebe:
Oops, missed a needed commit

PR #148, opened 2018-12-10
Fixes for chisel-testers add_enum_support

ucbjrl:
Update DspTester signatures to avoid conflicts with new chisel-testers enum support.


ucbjrl:
I'll push new SNAPSHOTs then update the .travis.yml to build from SNAPSHOTS (since we now publish them).

PR #155, opened 2019-03-12
Rocket only builds for Scala 2.12

grebe:


grebe:
Oops, I had a stale master and didn't realize. My bad.

ucbjrl:
But why didn't this appear as out of date? The diff didn't match what was currently in master.

grebe:
I think I pulled master before making my change, and didn't realize that the pull updated build.sbt. I was just being dumb, I think

PR #157, opened 2019-04-25
Make DspComplex multiplies use growing addition.

grebe:
DspComplex relies on using the underlying Ring's multiplication. For UInt, SInt, FixedPoint, etc. these additions wrap instead of grow. For our implementation of complex multiplication, growing is more desirable
than wrapping.

This commit adds specialized versions of the DspComplex for UInt, SInt, FixedPoint, etc. that use growing addition.

This is in response to an issue @mkosunen and @yuedai14 were running into that they were working around by increasing input bitwidth.

@shunshou, what do you think?

### DIFF
--- a/src/main/scala/dsptools/numbers/chisel_types/DspComplexTypeClass.scala
+++ b/src/main/scala/dsptools/numbers/chisel_types/DspComplexTypeClass.scala
@@ -3,18 +3,30 @@
 package dsptools.numbers
 
 import chisel3._
+import chisel3.experimental.FixedPoint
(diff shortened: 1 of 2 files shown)


shunshou:
Fine by me. It would be nice if Firrtl could throw a warning when the
output of some chain of multiplications grew to have too many bits (sanity
check in case people generated waste fully large intermediate terms)

shunshou:
Hm actually my recommendation was always to use context_(op) and have
people explicitly define behavior.

PR #163, opened 2019-06-06
Fix standaloneParams for TL

grebe:


grebe:
Ugh, I confused myself lol

PR #168, opened 2019-06-25
Support stage/phase OptionsException wrapping.

ucbjrl:
Test the exception cause in the "UInt subtract with overflow type Grow not supported" test (since stage/phase code will wrap it in an OptionsException).


ucbjrl:
This address freechipsproject/chisel3#1114.

PR #172, opened 2019-08-14
Ensure Driver.execute returns true.

ucbjrl:
`MemMasterSpec.should work with TileLink` was passing with an invalid backend apparently due to the fact that the `false` returned by `Driver.execute` wasn't triggering a failure. The ScalaTest documentation for [FlatSpec](http://doc.scalatest.org/1.1/org/scalatest/FlatSpec.html) would seem to indicate that in the absence of exceptions, assertions should be used to verify behavior.

### DIFF
--- a/rocket/src/test/scala/tester/MemMasterSpec.scala
+++ b/rocket/src/test/scala/tester/MemMasterSpec.scala
@@ -129,7 +129,10 @@ class MemMasterSpec extends FlatSpec with Matchers {
 
   behavior of "MemMaster Tester"
 
-  it should "work with TileLink" in {
+  // The following test is ignored, since if currently (8/23/19) fails with:
+  //  [info] [0.008] Assertion failed: 'A' channel Get carries invalid source ID (connected at MemMasterSpec.scala:35:8)
+  //  [info] [0.009]     at Monitor.scala:73 assert (source_ok, "'A' channel Get carries invalid source ID" + extra)
+  it should "work with TileLink" ignore {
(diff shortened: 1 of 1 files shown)


ucbjrl:
I've fixed the check failures (updated chisel-testers hadn't been published). @grebe could you review this now?

rocket/src/test/scala/tester/MemMasterSpec.scala
```diff
@@ -129,25 +129,31 @@ class MemMasterSpec extends FlatSpec with Matchers {
...
+  // The following test is ignored, since it currently (8/23/19) fails with:
+  //  [info] [0.008] Assertion failed: 'A' channel Get carries invalid source ID (connected at MemMasterSpec.scala:35:8)
+  //  [info] [0.009]     at Monitor.scala:73 assert (source_ok, "'A' channel Get carries invalid source ID" + extra)
+  it should "work with TileLink" ignore {
     lazy val dut = LazyModule(new TLRegmapExample)
     // use verilog b/c of verilog blackboxes in TileLink things
-    chisel3.iotesters.Driver.execute(Array[String]("-tbn", "verilator"), () => dut.module) { c =>
+    val result = chisel3.iotesters.Driver.execute(Array[String]("-tbn", "verilator"), () => dut.module) { c =>
       new TLRegmapExampleTester(dut)
     }
```
grebe: I think the more canonical way to do this is (unless @chick thinks differently)
```
chisel3.iotesters.Driver.execute(...) { c => 
 new WhateverTester(c)
} should be (true)
```

PR #173, opened 2019-08-20
Support migration of dontTouch, RawModule, MultiIOModule from experimental

ucbjrl:
This depends on freechipsproject/chisel3#1162


rocket/src/test/scala/tester/MemMasterSpec.scala
```diff
@@ -129,7 +128,7 @@ class MemMasterSpec extends FlatSpec with Matchers {
 
   behavior of "MemMaster Tester"
 
-  it should "work with TileLink" in {
+  it should "work with TileLink" ignore {
```
chick: maybe a comment should be added about why this is switched to ignore

PR #176, opened 2019-10-23
Fixed an enable signal bug for single delay shift registers.

milovanovic:
A minor bugfix in single delay shift registers that are now correctly implemented with an enable signal, accompanied by additional tests.

### DIFF
--- a/rocket/src/main/scala/craft/ShiftRegisterMem.scala
+++ b/rocket/src/main/scala/craft/ShiftRegisterMem.scala
@@ -17,8 +17,7 @@ object ShiftRegisterMem {
     if (n == 0) {
       in
     } else if (n == 1) {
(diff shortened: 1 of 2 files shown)


grebe:
Sorry for the slow reply.

LGTM. @chick, what do you think?

rocket/src/main/scala/craft/ShiftRegisterMem.scala
```diff
@@ -17,8 +17,7 @@ object ShiftRegisterMem {
     if (n == 0) {
       in
     } else if (n == 1) {
-      val out = Reg(in.cloneType)
-      out := in
+      val out = RegEnable(in, en)
```
chick: This is basically ok but I would prefer to see the parameters labelled when variable names don't provide
usage context

PR #177, opened 2019-11-05
Bump master SNAPSHOT version.

ucbjrl:
Bump master version number for next major release cycle.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,9 +34,8 @@ def javacOptionsVersion(scalaVersion: String): Seq[String] = {
 
 // Provide a managed dependency on X if -DXVersion="" is supplied on the command line.
 val defaultVersions = Map(
-  "chisel3" -> "3.3-SNAPSHOT",
   "chisel-iotesters" -> "1.4-SNAPSHOT",
-  "rocketchip" -> "1.3-SNAPSHOT"
+  "rocketchip" -> "1.2-SNAPSHOT"
 )
 
 name := "dsptools"
--- a/build.sbt
+++ b/build.sbt
@@ -109,9 +109,9 @@ val dsptoolsSettings = Seq(
 
 val rocketSettings = Seq(
     name := "rocket-dsptools",
-    libraryDependencies ++= Seq("chisel3", "chisel-iotesters", "rocketchip").map {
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -34,16 +34,16 @@ def javacOptionsVersion(scalaVersion: String): Seq[String] = {
 
 // Provide a managed dependency on X if -DXVersion="" is supplied on the command line.
 val defaultVersions = Map(
-  "chisel3" -> "3.2-SNAPSHOT",
-  "chisel-iotesters" -> "1.3-SNAPSHOT",
-  "rocketchip" -> "1.2-SNAPSHOT"
+  "chisel3" -> "3.3-SNAPSHOT",
```
chick: Does this need to be here, doesn't it get the right chisel from the testers dependency

ucbjrl: It should (get the correct chisel3 version from the testers' dependency). I think it's more of a "documentation" issue than an implementation one. We certainly could delete this. It would make it easier for experts to manage their dependencies. Let's see what @grebe thinks.

grebe: I feel like it's probably better to delete.

build.sbt
```diff
@@ -34,16 +34,16 @@ def javacOptionsVersion(scalaVersion: String): Seq[String] = {
...
+  "chisel-iotesters" -> "1.4-SNAPSHOT",
+  "rocketchip" -> "1.3-SNAPSHOT"
```
grebe: I don't think rocket has been bumped

ucbjrl:
I backed out the rocket-chip version bump, but this poses problems if we don't bump it. I don't think we can use a single version for rocket-chip if we want to publish one SNAPSHOT that's compatible with the 3.2.x/1.2.x branch series and a second SNAPSHOT that's compatible with the 3.3.x/1.3.x (i.e., current master) series.

grebe:
Can rocket now be used on the release branch or chisel? I know the `experimental.IO()` stuff used to make that impossible, although I think that's been released.

ucbjrl:
rocket-chip currently builds with the master branches of chisel and firrtl.

ucbjrl:
I'm waiting for the 3.3/1.3 snapshots to be published, before merging this PR.

PR #178, opened 2019-11-06
Bump deps, and clean up some imports

grebe:
Java 13 isn't cooperating. @ucbjrl, does this look OK?

### DIFF
--- a/rocket/src/main/scala/craft/ShiftRegisterMem.scala
+++ b/rocket/src/main/scala/craft/ShiftRegisterMem.scala
@@ -3,7 +3,7 @@
 package craft
 
 import chisel3._
-import chisel3.core.requireIsHardware
+import chisel3.internal.requireIsHardware
 import chisel3.util._
 
 object ShiftRegisterMem {
--- a/rocket/src/main/scala/dspblocks/MemToRegmap.scala
+++ b/rocket/src/main/scala/dspblocks/MemToRegmap.scala
@@ -1,7 +1,7 @@
 package dspblocks
 
 import chisel3._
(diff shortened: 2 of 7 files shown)


rocket/src/main/scala/craft/ShiftRegisterMem.scala
```diff
@@ -3,7 +3,7 @@
 package craft
 
 import chisel3._
-import chisel3.core.requireIsHardware
+import chisel3.internal.requireIsHardware
```
ucbjrl: Shouldn't this be `import chisel3.experimental.requireIsHardware`?

rocket/src/main/scala/dspblocks/MemToRegmap.scala
```diff
@@ -1,7 +1,7 @@
 package dspblocks
 
 import chisel3._
-import chisel3.core.requireIsChiselType
+import chisel3.internal.requireIsChiselType
```
ucbjrl: Shouldn't this be `import chisel3.experimental.requireIsChiselType`?

rocket/src/main/scala/craft/ShiftRegisterMem.scala
grebe: Should it? This is what the deprecation warning chisel3.core gave me. I'll switch if experimental is preferred

ucbjrl: Hmm. Maybe I'm out of step. The `experimental` piece may be a compatibility bone. If the warning says to use `internal`, then I'd use it.

PR #182, opened 2019-11-29
Add a set of deterministic rounding modes.

milovanovic:
Added a set of various deterministic rounding modes and the appropriate tests.

### DIFF
--- a/src/main/scala/dsptools/numbers/DspContext.scala
+++ b/src/main/scala/dsptools/numbers/DspContext.scala
@@ -14,25 +14,21 @@ case object Grow extends OverflowType
 abstract class TrimType
 case object Truncate extends TrimType
 case object NoTrim extends TrimType
-case object Floor extends TrimType
 case object RoundDown extends TrimType
-case object Ceiling extends TrimType
 case object RoundUp extends TrimType
 case object RoundTowardsZero extends TrimType
 case object RoundTowardsInfinity extends TrimType
 case object RoundHalfDown extends TrimType
 case object RoundHalfUp extends TrimType
 case object RoundHalfTowardsZero extends TrimType
-case object Round extends TrimType
 case object RoundHalfTowardsInfinity extends TrimType
-case object Convergent extends TrimType
 case object RoundHalfToEven extends TrimType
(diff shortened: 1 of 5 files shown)


src/main/scala/dsptools/numbers/DspContext.scala
```diff
@@ -13,12 +13,24 @@ case object Grow extends OverflowType
 /** Different trim methods */
 abstract class TrimType
 case object Truncate extends TrimType
-case object RoundHalfUp extends TrimType
 case object NoTrim extends TrimType
 case object Floor extends TrimType
+case object RoundDown extends TrimType
+case object Ceiling extends TrimType
+case object RoundUp extends TrimType
```
grebe: Is there a reason to have both `Ceiling` and `RoundUp`? I believe they have the same behavior. I think it would be better to have just one. If you want another name for the same thing for clarity purposes, I think a val in a package object acting as an alias would be the way to go.

milovanovic: You are absolutely right. Indeed, according to [Wikipedia](https://en.wikipedia.org/wiki/Rounding "Wikipedia's article on Rounding") and many other sources these two (`Ceiling` and `RoundUp`) are synonyms, just as `Floor` and `RoundDown`, or `Convergent` and `RoundHalfToEven` are. Since both naming conventions are present in everyday use, it seemed natural to have both cases supported inside the pattern matching, but identical expression triggered by each of the two. If you believe keeping only one and discarding the other option is better, feel free to delete one of the two that you think is less common. You're a native speaker, after all.

grebe: I created a branch [here](https://github.com/grebe/dsptools/tree/roundingmodes) where I picked some names- if you disagree, let me know! If you merge that branch in I'll push.

milovanovic: Here's an implementation of rounding aliases through already existing package object. If for some reason this isn't satisfactory, we can go without aliases.

grebe:
The test that is failing is waiting on an upstream fix, so I'm going to merge this despite the failure

PR #185, opened 2019-12-19
Add (newly) expected name argument to PeripheryBus constructor.

ucbjrl:
As of chipsalliance/rocket-chip#2229, the `PeripheryBus()` constructor has grown a new `String` parameter (without a default). Add it to the call in `DmaSpec` tests.


ucbjrl:
Superseded by #203

PR #194, opened 2020-02-11
Use the same version of verilator as firrtl

grebe:


grebe:
@ucbjrl do you know why travis is dying?

ucbjrl:
In their infinite wisdom, the verilator admins appear to have deleted the old tags for verilator versions. Try using `v3.922` as a tag.

PR #211, opened 2020-08-02
Fft revival 2

chick:
Brings over many function enhancements centered on production of an FFT DspBlock
Add dspmath package …
Add BitWidth IAUtility …
Update BinaryRepresentation
Update DspComplex
Update DspComplex and Node
Add BaseN
update dsptools.tester package

### DIFF
--- a/src/main/scala/dsptools/dspmath/Factorization.scala
+++ b/src/main/scala/dsptools/dspmath/Factorization.scala
@@ -1,0 +1,135 @@
+// See LICENSE for license details.
+
+package dsptools.dspmath
+
+import org.scalatest.{FlatSpec, Matchers}
+
+case class RadPow(rad: Int, pow: Int) {
+  /** `r ^ p` */
+  def get: Int = math.pow(rad, pow).toInt
+  /** Factorize i.e. rad = 4, pow = 3 -> Seq(4, 4, 4) */
+  def factorize: Seq[Int] = Seq.fill(pow)(rad)
+}
+
+case class Factorization(supportedRadsUnsorted: Seq[Seq[Int]]) {
+  /** Supported radices, MSD First */
+  private val supportedRads = supportedRadsUnsorted.map(_.sorted.reverse)
+
+  /** Factor n into powers of supported radices and store RadPow i.e. r^p, separated by coprimes
+    * i.e. if supportedRads = Seq(Seq(4, 2), Seq(3)),
+    * output =  Seq(Seq(RadPow(4, 5), RadPow(2, 1)), Seq(RadPow(3, 7)))
+    * implies n = 4^5 * 2^1 * 3^7
+    */
+  private def getRadPows(n: Int): Seq[Seq[RadPow]] = {
+    // Test if n can be factored by each of the supported radices (mod = 0)
+    // Count # of times it can be factored
+    var unfactorized = n
+    val radPows = for (primeGroup <- supportedRads) yield { for (rad <- primeGroup) yield {
+      var (mod, pow) = (0, 0)
+      while (mod == 0) {
+        mod = unfactorized % rad
+        if (mod == 0) {
+          pow = pow + 1
+          unfactorized = unfactorized / rad
+        }
+      }
+      RadPow(rad, pow)
+    }}
+    // If n hasn't completely been factorized, then an unsupported radix is required
+    require(unfactorized == 1, s"$n is invalid for supportedRads.")
+    radPows
+  }
+
+  /** Factor n into powers of supported radices (flattened)
+    * i.e. if supportedRads = Seq(Seq(4, 2), Seq(3)),
+    * output = Seq(5, 1, 7)
+    * implies `n = 4^5 * 2^1 * 3^7`
+    * If supportedRads contains more radices than the ones used, a power of 0 will be
+    * associated with the unused radices.
+    */
+  def getPowsFlat(n: Int): Seq[Int] = {
+    getRadPows(n).flatMap(_.map(_.pow))
+  }
+
+  /** Break n into coprimes i.e.
+    * n = 4^5 * 2^1 * 3^7
+    * would result in Seq(4^5 * 2^1, 3^7)
+    * If supportedRads contains more coprime groups than the ones used, 1 will be
+    * associated with the unused groups.
+    */
+  def getCoprimes(n: Int): Seq[Int] = {
+    getRadPows(n).map(_.map(_.get).product)
+  }
+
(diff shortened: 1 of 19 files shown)


src/main/scala/dsptools/numbers/binary_types/BinaryRepresentation.scala
```diff
@@ -20,5 +22,8 @@ trait BinaryRepresentation[A <: Data] extends Any {
   def mul2(a: A, n: Int): A = shl(a, n)
   // Trim to n fractional bits (with DspContext) -- doesn't affect DspReal
   def trimBinary(a: A, n: Int): A = trimBinary(a, Some(n))
-  def trimBinary(a: A, n: Option[Int]): A 
+  def trimBinary(a: A, n: Option[Int]): A
+
+  // Clip A to B (range)
+  def clip(a: A, b: A): A = ???
```
grebe, line 28: I think the implementation (`???`) should not be here, but in whatever implements the trait

src/main/scala/dsptools/numbers/representations/MixedRadix.scala
```diff
@@ -0,0 +1,149 @@
...
+
+package dsptools.numbers.representations
+
+import org.scalatest.{FlatSpec, Matchers}
+import chisel3._
+import generatortools.io.CustomBundle
+
+
+object MixedRadix {
+  /** Convert a mixed-radix digit sequence (least significant digits highested indexed)
```
grebe, line 11:
```suggestion
  /** Convert a mixed-radix digit sequence (least significant digits highest indexed)
```

src/main/scala/dsptools/dspmath/Factorization.scala
```diff
@@ -0,0 +1,135 @@
+// See LICENSE for license details.
+
+package dsptools.dspmath
+
+import org.scalatest.{FlatSpec, Matchers}
+
+case class RadPow(rad: Int, pow: Int) {
+  /** `r ^ p` */
+  def get: Int = math.pow(rad, pow).toInt
```
grebe, line 9:
```suggestion
  def get: Int = BigInt(rad).pow(pow).toInt
```

grebe: I think this is correct, and better?

src/main/scala/generatortools/io/Utils.scala
```diff
@@ -0,0 +1,156 @@
...
+object ConvertType {
+  /** Convert special type (i.e. Clock, Interval) to equivalent base Chisel type.
+    * Otherwise, keep the current base Chisel type. Warning: Any kind of Record becomes a CustomBundle!
+    */
+  def apply[T <: Data](tpe: T): Data = {
+    val newTpe = tpe match {
+      // TODO: special case others
+      case s: SInt =>
+        s.widthOption match {
+          case None => SInt(64.W)
```
grebe, line 22: These rules all seem a bit scary and arbitrary. Is this actually used/necessary? It seems like a `cloneType` that wants to enforce having a width for some types, but not others. Maybe these places should be replaced with `???`

src/main/scala/generatortools/io/Utils.scala
```diff
@@ -0,0 +1,156 @@
...
+    case _ => (false, false)
+  }
+}
+
+object Connect {
+  /** Connect similar signals. Hierarchy should be identical (esp. for Records). This enables you to, for example,
+    * connect an Interval signal with an equivalent base Chisel signal. globalClk will be passed in to
+    * all clock inputs, if provided.
+    */
+  def apply[T <: Data, U <: Data](left: T, right: U, globalClk: Option[Clock] = None): Unit = {
```
grebe, line 120: This also seems a bit scary, and in this PR is only used in one place. Might be better to just do the connection by hand if there aren't other usages.

src/test/scala/generatortools/io/TestModule.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  if (name.nonEmpty) dut.suggestName(name)
+
+  // Make sure global clock/reset functionality doesn't accidentally get mixed in
+  if (useGlobalClk) dut.clock := clock
+  else dut.clock := false.B.asClock
+  if (useGlobalRst) dut.reset := reset
+  else dut.reset := true.B
+
+  Connect(io, dut.io, if (useGlobalClk) Some(clock) else None)
```
grebe, line 40:
```suggestion
io <> dut.io
```

grebe: Does this work? I remember bulk connect being really wonky in earlier chisel3 days, so maybe this was necessary, but I think `a <> b` should work when one is a `cloneType` of the other, right?

src/main/scala/dsptools/numbers/binary_types/BinaryRepresentation.scala
chick: What should the implementation be, or should I just throw an error for now. Making this method abstract requires implementing in DspComplex, DspReal, Interval, FixedPoint, SInt, UInt type classes

src/main/scala/dsptools/numbers/representations/MixedRadix.scala
chick: Done

src/main/scala/dsptools/numbers/binary_types/BinaryRepresentation.scala
grebe: It seems to me that all the inheritors should all have `???` until implemented so this one can be abstract. I think it's probably not good to have a default implementation for something like this because weird stuff can happen if you forget to override it. However, most of implementations can probably have something like

```
def clip(a: A, b: A): A = {
  requireIsHardware(a)
  requireIsHardware(b)
  WireDefault(b, a)
}
```

src/main/scala/generatortools/io/Utils.scala
chick: I'm going to take another crack at this this evening. I may pull this all out and up into fft-generator until it all gets tested and worked out

chick: I'm going to move the whole generatortools package up into fft-generator until it settles down

src/test/scala/generatortools/io/TestModule.scala
chick: Removed for now, we'll see how it evolves in fft-generator

chick:
@grebe sorry this took so long, but it got behind the chisel3 3.4.0-RC1 release effort. Tests here will not pass until a new snapshot release is published

PR #215, opened 2020-11-10
Update version and dependency versions

chick:
This is now 1.5-snapshot
depending on 1.6-snapshot of iotesters


jackkoenig:
This builds its dependencies locally and thus passes CI

jackkoenig (approved):

PR #224, opened 2021-03-23
Added JTAG-to-memory-mapped bus (TL & AXI4) master bridge.

milovanovic:
This adds a JTAG-to-memory-mapped bus master bridge, right now only for TL and AXI4, but could be potentially extended to other buses like AHB or APB. The main idea is to be able to test (in hardware) memory-mapped peripherals (e.g., a `DspBlock`) by reading from and writing to their status and configuration registers without having to invoke a complete processor core like Rocket. Portions of the code are reused from [chisel-jtag](https://github.com/ucb-art/chisel-jtag) and/or [Rocket Chip JTAG](https://github.com/chipsalliance/rocket-chip/tree/master/src/main/scala/jtag). The same [PR](https://github.com/ucb-bar/testchipip/pull/120) is recently opened against [testchipip](https://github.com/ucb-bar/testchipip). Feel free to give improvement suggestions and to change the license to whatever you believe is the most suitable one.

PR #230, opened 2021-05-18
Compatibility updates 1

chick:


harrisonliew:
This supersedes #224?

chick:
I think this is the same as #224. Closing this

PR #235, opened 2021-07-14
Adds support for Scala 2.13

chick:
Adds support for Scala 2.13
- Fixes to build.sbt
  - Added 2.13.6 to cross build
  - Removed no longer needed Abe's workarounds of chiseltest/scalatest issues
  - Updated spire to a level that support 2.13
  - Removed java7 and 2.11 support stuff
- Added DoubleRange to deal with 2.13 not supporting Range constructed with doubles
  - Many changes to tests that used double ranges
  - Test for DoubleRange


src/test/scala/dsptools/DspContextSpec.scala
```diff
@@ -7,6 +7,11 @@ import dsptools.numbers._
 import org.scalatest.freespec.AnyFreeSpec
 import org.scalatest.matchers.should.Matchers
 
+// The following line is necessary for running the parallel test section but is not compatible
+// with 2.12.
+//import scala.collection.parallel.CollectionConverters.RangeIsParallelizable
+//
```
jackkoenig, line 13: Please don't just add commented code. If you really need this import, then you can add a file with a shim object that doesn't do anything, see `SeqMap` in chisel3 which is provided as an alias to `ListMap` for Scala 2.12: https://github.com/chipsalliance/chisel3/blob/master/core/src/main/scala-2.12/scala/collection/immutable/package.scala

I'd just add the following to a file `src/main/scala-2.12/scala/collection/parallel/CollectionConverters.scala`:
```scala
package scala.collection.parallel

object CollectionConverters {
  object RangeIsParallelizable
}
```

src/main/scala/dsptools/numbers/DoubleRange.scala
```diff
@@ -0,0 +1,44 @@
...
+  def apply(start: Double, stop: Double, step: Double = 1.0): DoubleRange = {
+    DoubleRange(start, stop, step)
+  }
+}
+
+object DoubleRangeUntil {
+  def apply(start: Double, stop: Double, step: Double = 1.0): DoubleRange = {
+    DoubleRange(start, stop, step, isTo = false)
+  }
+}
```
jackkoenig, line 44: I would not bother adding this type (and the tests), instead I'd just just use `BigDecimal(start) to <stop> by <step>` in the code

jackkoenig (commented):
Looks good for the most part, just some thoughts on the `DoubleRange` stuff

chick:
@jackkoenig Implemented suggested changes
- 2.12 specific shim did the trick
- not entirely happy with BigDecimal stuff, but problem is deeper with the overall code's ability to use BigDecimal. We should have used that all along but I guess not too many people are building circuits with fixed or floating point numbers bigger than 64 bits

jackkoenig (approved):

PR #236, opened 2021-09-18
Ignore tests that put literals in Bundle types

jackkoenig:
This was never intended to be supported in Chisel3 and is more
explicitly banned in Chisel v3.5.0-RC1. These tests can become compliant
with the stricter API by using Bundle literals rather than Bundles with
literals for fields.

The check these are failing was introduced in https://github.com/chipsalliance/chisel3/pull/2046.

I think these tests should be fixed but I don't want to block the release candidate on them.

Annoyingly, these are really hard to debug because the stack trace trimming is hiding everything which is a separate issue that we should address, here's a useful diff to apply to help when debugging this:
```diff
diff --git a/src/main/scala/dsptools/Driver.scala b/src/main/scala/dsptools/Driver.scala
index e776362..4ae6da2 100644
--- a/src/main/scala/dsptools/Driver.scala
+++ b/src/main/scala/dsptools/Driver.scala
@@ -31,6 +31,9 @@ object Driver {
       optionsManager.treadleOptions = optionsManager.treadleOptions.copy(
         blackBoxFactories = optionsManager.treadleOptions.blackBoxFactories :+ new TreadleDspRealFactory
       )
+      optionsManager.chiselOptions = optionsManager.chiselOptions.copy(
+        printFullStackTrace = true
+      )
       iotesters.Driver.execute(dutGenerator, optionsManager)(testerGen)
     }
 
```

### DIFF
--- a/src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
+++ b/src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala
@@ -366,7 +366,7 @@ class SimpleTBSpec extends AnyFlatSpec with Matchers {
 
   behavior of "simple module lits"
 
-  it should "properly read lits with gen = sint (reals rounded) and expect tolerance set to 1 bit" in {
+  ignore should "properly read lits with gen = sint (reals rounded) and expect tolerance set to 1 bit" in {
     val opt = new DspTesterOptionsManager {
       dspTesterOptions = optionsPass.dspTesterOptions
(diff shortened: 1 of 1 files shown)


jackkoenig:
I believe the error stems from Bundles like this one https://github.com/ucb-bar/dsptools/blob/b727fcdf6f8b204cb5bbbb8c005e97a0f71d54f0/src/test/scala/dsptools/VerboseDspTesterSpec/SimpleTBwGenTypeOption.scala#L158 that need to be turned into Bundle lits

PR #241, opened 2021-12-17
Fix multiple issues for 3.5.0-RC2 Release

jackkoenig:
This ballooned into several fixes in order to get CI passing:
* Use iotesters polyfill for removed Logger method
* Use chisel3 compiler plugin and delete cloneType implementations
* Fix DspComplex handling of lits
* Fix DspReal handling of lits 

In the case of `DspReal`, this is unfortunately a source incompatible change. It's not clear if it's possible to do this in a way that would be backwards compatible. The previous code was including a literal in the elements of the `DspReal` `Bundle` and this is something that was never really intended in Chisel3. It violates many internal assumptions and often would result in errors. The checking for this case now happens earlier and thus the code had to change.

Needed for 3.5.0-RC2

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,6 +6,7 @@ enablePlugins(GhpagesPlugin)
 
 val defaultVersions = Map(
   "chisel-iotesters" -> "2.5-SNAPSHOT",
+  "chisel3" -> "3.5-SNAPSHOT",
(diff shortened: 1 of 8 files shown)


chick (approved):
LGTM

PR #244, opened 2022-12-30
Remove Intervals.

milovanovic:
This removes `Interval` data type from dsptools in  an overall effort to keep pace with the future Chisel versions in which it'll be deprecated.


jerryz123 (approved):

jerryz123:
Thanks for the work here, this is much appreciated

abejgonzalez (approved):

PR #248, opened 2023-08-17
Reintroduce tests that use chiseltest.iotesters and API clean-up

milovanovic:
* Reintroduces some of the tests that were deleted in the process of transition to Chisel 3.6
* Minor API clean-up
* Fixedpoint library update


jerryz123 (approved):

sequencer:
why not have a try to svsim?

konda-x1:
> why not have a try to svsim?

Pretty sure it's for Chisel 5 and above. Now that chiseltest also works with Chisel 5, and since it has the iotesters compatibility interface, it seemed appropriate to port the tests to chiseltest without too much overhead.

PR #251, opened 2024-01-04
Update Chisel version to 5.1.0

tymcauley:
Bumps `fixedpoint` to include https://github.com/ucb-bar/fixedpoint/pull/8.

Note that this is not compatible with Chisel 3.6.0.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,8 +5,8 @@ enablePlugins(SiteScaladocPlugin)
 enablePlugins(GhpagesPlugin)
 
 val defaultVersions = Map(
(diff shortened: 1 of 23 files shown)


jerryz123:
I'm actively investigating merging this. My plan is to proceed incrementally, we'll move to a system where most flows will run in chisel6, but more complex flows can rely on chisel3.6.

Obviously in such a system we'll need to reference multiple versions of projects such as this, to handle the backwards-compatibility-breaking changes in chisel6.
That will be temporary clutter until everything moves into chisel6.

